PR #10, opened 2019-09-18
Explain checks

mitchellhamilton:
@Noviny this is intended as both documentation for review which I'll also copy to the monorepo style guide and as context for an issue I'm about to open about a problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ The dep is removed from `devDependencies` or `optionalDependencies`.
 
 ### Invalid package name
 
-There are rules from npm about what a package name can be. This is already enforced by npm on publish but in a monorepo, everything will be published together so some packages may depend on a package which can't be published so we should check that package names are valid before publishing.
+There are rules from npm about what a package name can be. This is already enforced by npm on publish but in a multi-package repository, everything will be published together so some packages may depend on a package which can't be published. Checking for invalid package names prevents this kind of publish failure.
 
 ### Unsorted dependencies
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ yarn manypkg fix
 
 ## External mismatch
 
-The ranges for all dependencies(excluding `peerDependencies`) on external packages should be equal(`===`). This is so that only a single version of an external package will be installed because having multiple versions of the same package can cause confusion and bundle size problems especially with libraries like React that require there to only be a single copy of the library. It's important to note that this check does not enforce that only a single version of an external package is installed, only that two versions of an external package will never be installed because they're specified as dependencies of internal packages.
+The ranges for all dependencies(excluding `peerDependencies`) on external packages should exactly match(`===`). This is so that only a single version of an external package will be installed because having multiple versions of the same package can cause confusion and bundle size problems especially with libraries like React that require there to only be a single copy of the library. It's important to note that this check does not enforce that only a single version of an external package is installed, only that two versions of an external package will never be installed because they're specified as dependencies of internal packages.
 
 There are vague thoughts that it should be possible to partially disable external mismatch rules but we have not yet had a use case for it so we have not addressed it yet.
 
--- a/README.md
+++ b/README.md
@@ -20,34 +20,65 @@ yarn manypkg check
 yarn manypkg fix
 ```
 
-### Script Runner
+## Dictionary
 
-> Note: The following is not implemented currently
+- **private package** - A package that has `private: true`/is not published. It does not refer to a package published to a private registry here.
+- **internal package/workspace** - A package that is local/in the repo
+- **external package** - A package that is from a registry like npm
+- **range** - A [node-semver range](https://github.com/npm/node-semver#ranges)
 
-### In the root of a multi-package repo
+## Checks
 
-```
-yarn m <pkgName> add <depDescriptor>
-```
+## External mismatch
 
-```
-yarn m <pkgName> <scriptName>
-```
+The ranges for all dependencies(excluding `peerDependencies`) on external packages should be equal(`===`). This is so that only a single version of an external package will be installed because having multiple versions of the same package can cause confusion and bundle size problems especially with libraries like React that require there to only be a single copy of the library. It's important to note that this check does not enforce that only a single version of an external package is installed, only that two versions of an external package will never be installed because they're specified as dependencies of internal packages.
 
-```
-yarn m <binaryFromNodeModulesName>
-```
+There are vague thoughts that it should be possible to partially disable external mismatch rules but we have not yet had a use case for it so we have not addressed it yet.
 
-### In a package
+### How it's fixed
 
-```
-yarn m add <depDescriptor>
-```
+The highest range of the dependency is found as defined by [sembear](https://github.com/mitchellhamilton/sembear) and set as the range at every non-peer dependency place it is depended on.
 
-```
-yarn m <scriptName>
-```
+## Internal mismatch
 
-```
-yarn m <binaryFromNodeModulesName>
-```
+The ranges for all dependencies(excluding `peerDependencies`) on internal packages should include the version of the internal package. This is so that an internal package will never depend on another internal package but get the package from the registry because that happening is very confusing and you should always prefer a local version of any given package.
+
+### How it's fixed
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
 
-```
-yarn m <pkgName> <scriptName>
-```
+The ranges for all dependencies(excluding `peerDependencies`) on external packages should be equal(`===`). This is so that only a single version of an external package will be installed because having multiple versions of the same package can cause confusion and bundle size problems especially with libraries like React that require there to only be a single copy of the library. It's important to note that this check does not enforce that only a single version of an external package is installed, only that two versions of an external package will never be installed because they're specified as dependencies of internal packages.
 
-```
-yarn m <binaryFromNodeModulesName>
-```
+There are vague thoughts that it should be possible to partially disable external mismatch rules but we have not yet had a use case for it so we have not addressed it yet.
```
Noviny: I'd like to not have this in the main readme - I'd probably open an issue for discussion of this, or a `decisions` docs page that lists this.

> we have not yet had a use case for it

Slightly untrue - we have had people with use-cases who have either not adopted our tools, or begun work on their own equivalent - I think think this is worth having as an open (but possibly low priority) issue.

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
 
-```
-yarn m <pkgName> add <depDescriptor>
-```
+## External mismatch
 
-```
-yarn m <pkgName> <scriptName>
-```
+The ranges for all dependencies(excluding `peerDependencies`) on external packages should be equal(`===`). This is so that only a single version of an external package will be installed because having multiple versions of the same package can cause confusion and bundle size problems especially with libraries like React that require there to only be a single copy of the library. It's important to note that this check does not enforce that only a single version of an external package is installed, only that two versions of an external package will never be installed because they're specified as dependencies of internal packages.
```
Noviny: Change `should be equal` to `should exactly match` ?

(github suggestions were not making this legible - sorry for comment suggest change)

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
-```
+There are vague thoughts that it should be possible to partially disable external mismatch rules but we have not yet had a use case for it so we have not addressed it yet.
 
-### In a package
+### How it's fixed
 
-```
-yarn m add <depDescriptor>
-```
+The highest range of the dependency is found as defined by [sembear](https://github.com/mitchellhamilton/sembear) and set as the range at every non-peer dependency place it is depended on.
```
Noviny: I'd love if we could point to a dictionary entry as opposed to just 'as per this exeternal package' - particularly because nobody will instantly know about sembear (yet).

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+## Internal mismatch
 
-```
-yarn m <binaryFromNodeModulesName>
-```
+The ranges for all dependencies(excluding `peerDependencies`) on internal packages should include the version of the internal package. This is so that an internal package will never depend on another internal package but get the package from the registry because that happening is very confusing and you should always prefer a local version of any given package.
+
+### How it's fixed
+
+If the range is a [caret range](https://github.com/npm/node-semver#caret-ranges-123-025-004) or a [tilde range](https://github.com/npm/node-semver#tilde-ranges-123-12-1) with no other comparators, the range is set as a caret or tilde range respectively with the version of the internal package. If it is any other range, the range is set to the exact version of the internal package.
```
Noviny: > If it is any other range, the range is set to the exact version of the internal package.

This sentence needs more clarity - I *think* I know what this means (It will move `>=10.0.1` to `>=10.0.2` when the package gets patched) but I'm not sure if that's what it means.

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+All `peerDependencies` should also be specified in `devDependencies` and the range specified in `devDependencies` should be a subset of the range for that dependency in `peerDependencies`. This is so that `peerDependencies` are available in the package during development for testing and etc.
+
+### How it's fixed
+
+The range for the dependency specified in `peerDependencies` is added to `devDependencies` unless the package is already a non-peer dependency elsewhere in the repo in which, that range is used instead.
+
+### Root has devDependencies
+
+Whether a dependency is in `devDependencies` or `dependencies` do not make a difference in the root/private packages in general(though we don't enforce it for all private packages to support the case where a package is private for development or etc. and then published) so to avoid confusion as to where a root dependency should go, all dependencies should go in `dependencies`. This decision may be changed in the future to force all root dependencies to be in `dependencies` so that installation instructions for development packages do not have to be different for monorepos.
```
Noviny:
```suggestion
In the root `package.json` of a multi-package repository, whether a dependency is in `devDependencies` or `dependencies` does not make a difference. To avoid confusion as to where a root dependency should go, all dependencies should go in `dependencies`.
```

> This decision may be changed in the future to force all root dependencies to be in `dependencies` so that installation instructions for development packages do not have to be different for monorepos.

Aren't we already enforcing this with this rule?

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+#### How it's fixed
+
+All `devDependencies` at the root are moved to `dependencies`.
```
Noviny:
```suggestion
All `devDependencies` in the root `package.json` are moved to `dependencies`.
```

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+### Multiple dependency types
+
+A dependency shouldn't be specified in more than one of `dependencies`, `devDependencies` or `optionalDependencies` because what would that even do? TODO: find out what package managers actually do when this happens
```
Noviny: Make the TODO an issue?

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+#### How it's fixed
+
+The dep is removed from `devDependencies` or `optionalDependencies`.
```
Noviny: I assume this is if it is `dependencies` and either `devDependencies` or `optionalDependencies`. What happens if it's in `devDependencies` and `optionalDependencies`?

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+### Invalid package name
+
+There are rules from npm about what a package name can be. This is already enforced by npm on publish but in a monorepo, everything will be published together so some packages may depend on a package which can't be published so we should check that package names are valid before publishing.
+
```
Noviny:
```suggestion

#### How it's fixed

This requires manual fixing as automatically fixing this may lead to valid but incorrect package names.

```

Noviny: (added for consistency)

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+A dependency shouldn't be specified in more than one of `dependencies`, `devDependencies` or `optionalDependencies` because what would that even do? TODO: find out what package managers actually do when this happens
+
+#### How it's fixed
+
+The dep is removed from `devDependencies` or `optionalDependencies`.
+
+### Invalid package name
+
+There are rules from npm about what a package name can be. This is already enforced by npm on publish but in a monorepo, everything will be published together so some packages may depend on a package which can't be published so we should check that package names are valid before publishing.
```
Noviny:
```suggestion
There are rules from npm about what a package name can be. This is already enforced by npm on publish but in a multi-pakage repository, everything will be published together so some packages may depend on a package which can't be published. Checking for invalid package names prevents this kind of publish failure.
```

README.md
```diff
@@ -20,34 +20,65 @@ yarn manypkg check
...
+
+### Unsorted dependencies
+
+Dependencies should be sorted alphabetically because when you add a package with yarn add or etc. deps are sorted and that results in confusing diffs so we should enforce that they're sorted.
```
Noviny:
```suggestion
When you add a package with `yarn add` or etc. dependencies are sorted, and this can cause confusing diffs if the dependencies were not previously sorted.  Dependencies should be sorted alphabetically to avoid this.
```

README.md
mitchellhamilton: Whoops, I meant we change to devDependencies.

README.md
mitchellhamilton: Currently, we don’t handle that because I haven’t implemented it though the answer is probably optionalDependencies.

Noviny:
Merge this?

PR #17, opened 2019-10-03
look! A plugin

Noviny:
This is a night of hacking on this - it's in a usable state but with lots of 'ooh, not quite' in it.

Worth noting that we have an obvious use-case for this as a thing similar to here:

https://github.com/keystonejs/keystone-5/blob/master/website/gatsby-config.js

I want to push the filtering etc further down, into page creation. Also, for obvious reasons, keystone's creation rules aren't suitable for everyone so that would need to be config.

End goal of plugin-2 is that it generates a complete website section (default to `/packages`) displaying `README.md` and `CHANGELOG.md` for each package. Likely going to make it take a template.

Ah! I have so many thoughts about it, but I should sleep.

Okay, one more thought, which I'll X-post to other more general places: I'm inverting my thoughts around some things. I want a single gatsby plugin that I add, and it generates a whole website section, with the ability for me to add more common pages for each package. I think that's my end-goal with this. This section should be renderable in any gatsby site no worries.


Noviny:
Finally, this may be an inappropriate place to put this.

packages/gatsby-source-workspace/gatsby-node.js
```diff
@@ -0,0 +1,62 @@
...
+    `;
+  createTypes(typeDefs);
+};
+
+exports.sourceNodes = async (
+  { actions, createNodeId, options },
+  pluginOptions
+) => {
+  let { createNode } = actions;
+  let cwd = pluginOptions.cwd || (await badFindRepoRoot());
```
mitchellhamilton: lol, I wrote a thing to do this, it's in this repo, `find-workspaces-root`.

Noviny: :P I found it in my second round of editing on this (and implementing a slightly more opinionated extra layer)

Might move this to a draft PR

package.json
```diff
@@ -19,7 +20,7 @@
   },
   "preconstruct": {
     "packages": [
-      "packages/*"
+      "packages/[!gatsby]"
```
mitchellhamilton: TIL about this syntax, thanks!

packages/gatsby-monorepo-plugin/package.json
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "@manypkg/gatsby-monorepo-plugin",
```
mitchellhamilton: I think this is the convention

```suggestion
    "name": "@manypkg/gatsby-plugin-monorepo",
```

mitchellhamilton:
I'm really excited about this! 🎉 

> Finally, this may be an inappropriate place to put this.

I don't mind, my thinking on what manypkg is is that it's a bunch of things to help you in a monorepo so I think this makes sense here. (ofc if we want to move it later, we can do that)

package.json
Noviny: I got it wrong, it should be:

```
      "packages/[!gatsby]*"
```

Semi-related, if I wanted a list of packages preconstruct is taking charge of, how would I do that? Some form of `info` command.

(It's very low priority)

mitchellhamilton: That doesn't currently exist, I'm thinking of making a package to do that rather than a command, any objections to that?

Noviny: None at all other than curiousity as to why making it a separate package?

packages/gatsby-source-workspace/README.md
```diff
@@ -0,0 +1,16 @@
+# @manypgk/gatsby-source-workspace
```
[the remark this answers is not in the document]
Noviny: Varied and many expletives. Why is pacakge the word I most commonly misspell? How come I still misspell pacakge even when *I don't even write the damn word*?!

PR #24, opened 2019-10-23
Noviny/enforce internal dev dep range

### LINKED ISSUE #23: Update INTERNAL_MISMATCH rule for new devDep rule
We made a decisions that internal dev-dependencies should be `*` at all times, so that updating these does not cause unnecessary releases.

We need to update the internal mismatch rule to cover this.

Noviny:
closes #23


.changeset/cyan-cups-push.md
```diff
@@ -0,0 +1,31 @@
...
+{
+  "name": "sunshine",
+  "version": "1.0.0",
+  "devDepencies": {
+    "sun": "*"
+  }
+}
+```
+
+This is because all internal dependencies are always linked, and devDepencies are only relevant in local installs. Having set versions here caused packages to be patched when one of their devDepencies left the range, which was not strictly necessary.
```
mitchellhamilton:
```suggestion
This is because all internal dependencies are always linked if the version of theinternal dependency is within the specified range(which is already enforced by Manypkg), and devDependencies are only relevant in local installs. Having set versions here caused packages to be patched when one of their devDependencies left the range, which was not strictly necessary.
```

packages/cli/src/checks/INTERNAL_DEV_DEP_NOT_STAR.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  fix: error => {
+    let deps = error.workspace.config.devDependencies;
+    if (deps && deps[error.dependencyWorkspace.name]) {
+      deps[error.dependencyWorkspace.name] = "*";
+    }
+    return { requiresInstall: true };
+  },
+  print: error =>
+    `${error.workspace.name} has a dependency on ${error.dependencyWorkspace.name}@${error.dependencyRange} but the version of ${error.dependencyWorkspace.name} in the repo is ${error.dependencyWorkspace.config.version} which is not within range of the depended on version, please update the dependency version`,
```
mitchellhamilton: This message should be specific to this check

Noviny:
I am impressed how many times I made the exact same typo.

PR #42, opened 2020-02-13
Feature: Run command

tarang9211:


packages/cli/src/index.ts
```diff
@@ -81,11 +81,35 @@ async function execCmd(args: string[]) {
...
+    tools: ["yarn", "bolt", "pnpm", "root"]
+  }))!;
+  // console.log(workspaces);
+  const matchingWorkspaces = workspaces.filter(workspace => {
+    return workspace.name.includes(args[0]);
+  });
+
+  if (matchingWorkspaces.length > 1) {
+  } else if (matchingWorkspaces.length === 0) {
+    console.warn("No matching packages found");
```
mitchellhamilton:
```suggestion
    logger.error("No matching packages found");
```

packages/cli/src/index.ts
```diff
@@ -81,11 +81,35 @@ async function execCmd(args: string[]) {
...
+  let workspaces = (await getWorkspaces({
+    cwd: workspacesRoot,
+    tools: ["yarn", "bolt", "pnpm", "root"]
+  }))!;
+  // console.log(workspaces);
+  const matchingWorkspaces = workspaces.filter(workspace => {
+    return workspace.name.includes(args[0]);
+  });
+
+  if (matchingWorkspaces.length > 1) {
```
mitchellhamilton:
```suggestion
  if (matchingWorkspaces.length > 1) {
    logger.error(`an identifier must only match a single package but "${args[0]}" matches the following packages:\n${matchingWorkspaces.map(x=>x.name).join(\n)}`)
```

packages/cli/src/index.ts
```diff
@@ -81,11 +81,35 @@ async function execCmd(args: string[]) {
...
 
+async function runCmd(args: string[]) {
+  let workspacesRoot = await findWorkspacesRoot(process.cwd());
+  let workspaces = (await getWorkspaces({
+    cwd: workspacesRoot,
+    tools: ["yarn", "bolt", "pnpm", "root"]
+  }))!;
+  // console.log(workspaces);
+  const matchingWorkspaces = workspaces.filter(workspace => {
+    return workspace.name.includes(args[0]);
```
mitchellhamilton: Could you make this so you can match by the directory of the package as well?

PR #47, opened 2020-03-03
get-packages

tarang9211:


packages/get-packages/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "get-packages",
```
mitchellhamilton:
```suggestion
  "name": "@manypkg/get-packages",
```

packages/get-packages/README.md
```diff
@@ -0,0 +1,43 @@
...
+export type Package = { packageJson: PackageJSON; dir: string };
+
+type Packages = {
+  tool: Tool;
+  packages: Package[];
+  root: Package;
+};
+
+export function getPackages(cwd: string): Promise<Packages>;
+export function getPackagesSync(cwd: string): Packages;
```
Noviny: This info seems not ideal as this is not a normal way of explaining imports.

This should really just be the import statements. Having the tool, package, and packages types is useful.

Speaking of, can we expose these as exports? Having access to types is nice.

tarang9211: Do you want me to add exports to the types?

packages/get-packages/src/index.test.ts
```diff
@@ -0,0 +1,180 @@
+import fixturez from "fixturez";
+import { getPackages, getPackagesSync } from "./";
+
+const f = fixturez(__dirname);
+
+describe("getPackages", () => {
```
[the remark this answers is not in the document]
tarang9211: I think this is done, right?

PR #48, opened 2020-03-04
Add incorrect repository field check

mitchellhamilton:


mitchellhamilton:
@Noviny I'm gonna merge this now because I want it for a thing but feel free to leave comments

packages/cli/src/checks/INCORRECT_REPOSITORY_FIELD.ts
```diff
@@ -0,0 +1,71 @@
...
+export default makeCheck<ErrorType>({
+  type: "all",
+  validate: (workspace, allWorkspaces, rootWorkspace) => {
+    let rootRepositoryField = (rootWorkspace.config as any).repository;
+
+    if (typeof rootRepositoryField === "string") {
+      let result = parseGithubUrl(rootRepositoryField);
+      if (result !== null) {
+        if (workspace === rootWorkspace) {
+          let correctRepositoryField = `https://github.com/${result.owner}/${result.name}`;
```
Noviny: Just a thing, this makes manypkg exclusively for GH, which I don't think is ideal. Would love to make this more generic.

I'd also love a bit more flexibility in defining this, as this limits the spec (I personally quite like the

```
"repository": {
  "type" : "git",
  "url" : "https://github.com/facebook/react.git",
  "directory": "packages/react-dom"
}
```

format for this)

This comment really relates to further down, but would love to see if an object was used in the root, we recommended and fixed to an object.

Everything else looks good, some nice tidying done to tests and types.

mitchellhamilton: > Just a thing, this makes manypkg exclusively for GH, which I don't think is ideal. Would love to make this more generic.

💯. A little note with this: This doesn't make Manypkg exclusively for GH, it only makes this rule exclusive to GH

---

I specifically don't want to use objects because the directory field doesn't do anything as far as I can see (e.g. if you click the repository link on https://www.npmjs.com/package/react-dom, it goes to the root, not into the directory)

PR #53, opened 2020-03-10
#52: Add special logic to handle github repository separately from azure

### LINKED ISSUE #52: INCORRECT_REPOSITORY_FIELD requires github repos
My company doesn't use github, so we're seeing INCORRECT_REPOSITORY_FIELD errors with the new check.

This check should either be updated to handle other repositories, or at least disabled for non-github URLs.

NateRadebaugh:
Fixes #52 

Add special logic to handle github repository separately from azure.

I'm not sure if this is the most elegant solution, but it seems to work for me.


NateRadebaugh:
This PR doesn't handle things perfectly... in Azure DevOps, there can be multiple repositories:
* `https://dev.azure.com/<tenant>/<project>/_git/<repository>`

packages/cli/src/checks/INCORRECT_REPOSITORY_FIELD.ts
```diff
@@ -18,8 +18,15 @@ export default makeCheck<ErrorType>({
     if (typeof rootRepositoryField === "string") {
       let result = parseGithubUrl(rootRepositoryField);
       if (result !== null) {
```
mitchellhamilton:
```suggestion
      if (result !== null && (result.host === 'github.com' || result.host === 'dev.azure.com')) {
```

NateRadebaugh: Thanks, changed!

NateRadebaugh:
Thanks for the review @mitchellhamilton, I've updated the PR with your fixes. Anything else you'd like me to change or discuss?

PR #63, opened 2020-05-11
feat: Add support for Lerna monorepos

evocateur:
Howdy! 👋 

I like the level of abstraction of these packages, and wanted to make them useful for folks using Lerna monorepo patterns.

Thanks!

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -115,10 +115,13 @@ async function execCmd(args: string[]) {
         {
           yarn: "yarn",
           pnpm: "pnpm",
+          lerna: "lerna",
           root: "yarn",
           bolt: "bolt"
         }[tool],
-        tool === "pnpm" ? ["install"] : [],
+          tool === "pnpm" ? ["install"]
+        : tool === "lerna" ? ["bootstrap", "--since", "HEAD"]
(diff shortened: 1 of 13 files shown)


packages/cli/src/index.ts
```diff
@@ -115,10 +115,13 @@ async function execCmd(args: string[]) {
         {
           yarn: "yarn",
           pnpm: "pnpm",
+          lerna: "lerna",
           root: "yarn",
           bolt: "bolt"
         }[tool],
-        tool === "pnpm" ? ["install"] : [],
+          tool === "pnpm" ? ["install"]
+        : tool === "lerna" ? ["bootstrap", "--since", "HEAD"]
```
mitchellhamilton, line 123: What does `--since HEAD` do?

evocateur:
It ensures that bootstrap will only run against the packages changed by the fixes (and their local direct dependents) instead of the entire monorepo, in most cases.

packages/cli/src/index.ts
evocateur: It ensures that bootstrap will only run against the packages changed by the fixes (and their local direct dependents) instead of the entire monorepo, in most cases.

PR #66, opened 2020-05-17
feat: find Rust Cargo.toml with backward compatibility

jbolda:
This PR adds functionality to find both JavaScript based packages (as before), and Rust. I added a 'toolset' flag that requires opting in for the new functionality to maintain backwards compatibility if someone is using this in a js/rust polyglot repo and isn't prepared to start supporting Rust yet. This also means that a repo may now have multiple "roots", effectively one per language.

I see that many of the types are exported. Are these actually used anywhere else? I made a handful of the types that were packageJson specific into a more generic form, and used that. If they _are_ used elsewhere, I will need to add them back purely for the export to maintain backwards compatibility.

Opening this as a draft, that I can get a few eyes on it from @Noviny, @mitchellhamilton, @Andarist, and others that it still maintains the spirit of what you are after. The following are items that to be attended to in this PR. Note, there are a few tests that error incorrectly on windows as the `/` vs `\` resolves differently.

- [x] add rust to find-root
- [x] add rust to get-packages
- [x] pass first group of tests
- [x] resolve existing type errors and some new type errors
- [ ] merge in master and implement new features (e.g. lerna support
- [ ] resolve remaining type errors
- [ ] add a few more tests perhaps? any recommendations on edge cases to check?

refs:
https://github.com/jbolda/changesets/issues/1 notes on working on PR that will incorporate these changes eventually.
https://github.com/atlassian/changesets/issues/310 changes originally proposed in this issue, and then this package was extracted out to this repo.

### DIFF
--- a/.changeset/rude-tables-press.md
+++ b/.changeset/rude-tables-press.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/find-root": patch
+---
+
+add ability to find root of Rust repo
--- a/.changeset/rude-tables-press.md
+++ b/.changeset/rude-tables-press.md
@@ -1,5 +1,6 @@
 ---
 "@manypkg/find-root": patch
+"@manypkg/get-packages": patch
 ---
 
 add ability to find root of Rust repo
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ dist/
 .cache
 public
 scratchings.js
+
+# rust RLS adds these, not needed in fixtures
+__fixtures__/**/target
+__fixtures__/**/Cargo.lock
--- a/__fixtures__/rust-and-js/Cargo.toml
+++ b/__fixtures__/rust-and-js/Cargo.toml
@@ -1,0 +1,3 @@
+[package]
+name = "rust-and-js-crate-fixture"
+version = "0.5.0"
--- a/__fixtures__/rust-and-js/package.json
+++ b/__fixtures__/rust-and-js/package.json
@@ -1,0 +1,4 @@
+{
+  "name": "rust-and-js-package-fixture",
+  "version": "1.0.0"
+}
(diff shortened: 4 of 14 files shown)


Andarist:
The use case of managing polyglot (or even just non-js) projects with Changesets is very cool. I'm wondering though if baking support for non-js tools into the core functionality is a good idea. I suspect that many languages don't even have a standard way for defining monorepos (very nice that Rust have this!) but ideally, they should be able to use Changesets as well.

Baking this into the core also raises maintainability costs & overhead as core contributors are most likely not using Rust (or other tools) that often. 

Wouldn't it be possible to pursue an alternative solution for this that would focus on abstracting some interfaces and allowing "plugins"/overrides for things like `get-packages`? That way people could integrate Changesets with just anything on their own. A curated list of integrations could be kept but they would be maintained by people most interested in those tools.

Andarist:
> I think the more annoying bit for myself was maintaining both the sync and async functions, but don't know that there is a way to improve that situation.

[This package](https://github.com/loganfsmyth/gensync/) could be potentially used to help with this.

jbolda:
Oh, that is slick. I was wondering about generators myself as the solution. Support is good in node LTS as well. I suspect that is outside of the scope of this PR though.

jbolda:
The more that I think about this, the more that I think it makes sense to split this up into separate functions for each language. I think I will let everyone take a gander for a little longer on the current iteration of code, and then extract things out if there is no opposition.

mitchellhamilton:
Hey folks, @Noviny and I have some thoughts on this and the related issue, I've put it in the issue on Changesets to keep it in one place - https://github.com/atlassian/changesets/issues/310#issuecomment-630540105

jbolda:
Closing this put discussion in aforementioned issue.

PR #67, opened 2020-06-25
Add --default-branch cli switch for INCORRECT_REPOSITORY_FIELD check

jesstelford:
My repositories use `main` as the default branch, which causes `manypkg check` to incorrectly error with:

```
$ yarn manypkg check
...
☔️ error @example/button has a repository field of "https://github.com/example/repo/tree/main/button" when it should be "https://github.com/example/repo/tree/master/button"
error Command failed with exit code 1.
```

With this new cli swich, I can successfully execute:

```
$ yarn manypkg check --default-branch=main
...
✨  Done in 0.15s.
```

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,9 +1,9 @@
 // @flow
 import * as logger from "./logger";
 import fs from "fs-extra";
-import { getPackages, Package } from "@manypkg/get-packages";
+import { getPackages, Packages, Package } from "@manypkg/get-packages";
 import path from "path";
-import { Check } from "./checks/utils";
+import { Check, Options } from "./checks/utils";
 import { checks } from "./checks";
 import { ExitError } from "./errors";
 import { writePackage } from "./utils";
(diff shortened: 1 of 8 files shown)


packages/cli/src/index.ts
```diff
@@ -102,7 +139,8 @@ async function execCmd(args: string[]) {
   let { hasErrored, requiresInstall } = runChecks(
     packagesByName,
     root,
-    shouldFix
+    shouldFix,
+    cli.flags
```
jesstelford, line 143: By doing it this way, I figure it can be enhanced with something like `cosmiconfig` in the future to read these kinds of options from a file.

jesstelford:
Done - turned out fairly easy other than the TypeScript type trickery needed to add the options:

Add package.json#manypkg config object:

```
{
  "manypkg": {}
}
```

To support setting a default branch for the INCORRECT_REPOSITORY_FIELD check/fix, a new config option can be set:

```
{
  "manypkg": {
    "defaultBranch": "master"
  }
}
```

The default `defaultBranch` is `"master"`.

PR #68, opened 2020-07-12
Ignore invalid ranges for external packages

### LINKED ISSUE #37: Can't specify URL Tarball, instead of semver range, for dependencies
## Background 
We have a dependency entry in our root package.json, and package-a/package.json, that directly links to a tarball, e.g.:
```
{
  ...
  "dependencies": {
    ...
    "bootstrap": "https://github.com/twbs/bootstrap/archive/v3.3.5.tar.gz"
    ...
  }
}
```

## Actual Behaviour
When running `yarn manypkg check` we see it fail with the following error

```
☔️ error TypeError: Invalid comparator: https://github.com/twbs/bootstrap/archive/v3.3.5.tar.gz
☔️ error     at Comparator.parse (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:794:11)
☔️ error     at new Comparator (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:777:8)
☔️ error     at Range.<anonymous> (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:975:12)
☔️ error     at Array.map (<anonymous>)
☔️ error     at Range.parseRange (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:974:13)
☔️ error     at Range.<anonymous> (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:917:17)
☔️ error     at Array.map (<anonymous>)
☔️ error     at new Range (/my-monorepo/node_modules/sembear/node_modules/semver/semver.js:916:40)
☔️ error     at upperBound (/my-monorepo/node_modules/sembear/dist/sembear.cjs.dev.js:19:11)
☔️ error     at /my-monorepo/node_modules/sembear/dist/sembear.cjs.dev.js:138:19
```

## Expected Behaviour
manypkg check should gladly accept [tarball urls](https://docs.npmjs.com/files/package.json#urls-as-dependencies) if they are consistent across the package.jsons in the repo.

In the case where there is a mix of tarball urls or semver ranges, it should return a warning that it couldn't reconcile the versions in the repo.

## Hypothesis
It looks like sembear is assuming that it will be able to coerce the range into a valid value. I'm guessing we should do a check as to whether a dependency version is valid semver or not, and avoid trying to coerce it.

### LINKED ISSUE #65: sembear fails when trying to compare a git version type
When a package dep references github directly, the checker fails. I believe its got something to do with:

https://github.com/Thinkmill/manypkg/blob/a4db72a8b272f1b642fa751639d7840f4fa3658c/packages/cli/src/checks/utils.ts#L112

That being said - I have started to raise a pr that simply ignores `git:` but then it dawned on me that this'll also be an issue for `file:` etc... So I'm opening up the issue to discuss a best approach.

I believe that perhaps we can simply wrap that `highest` call in a try catch, so if it throws we could simply move on? Perhaps there is a way to first detect if its a semver version string, and if so only then check it.

I do accept that this is probably not best practice in using git, file and so on - references but there have been times where I'd want to still have the level of strictness manypkg offer, but not semever. Also I'd say that this wouldnt be verifiying that the strings are greater or less than - perhaps manypkg could error if the strings it finds aren't equal, ie: `"packageA": "git://github.com/abc/xyz.git#master"` is not the same as `"packageA": "git://github.com/abc/xyz.git#random-hash"`.

mitchellhamilton:
Closes #65
Closes #37


mitchellhamilton:
> it's really a workaround right?

Yes

> What if someone isn't installing from npm?

As in something like a git dependency? Those things would also be ignored here(hence why it closes #65 and #37). Or do you mean another registry that isn't _the_ npm registry? I'm 99% sure `npm:` doesn't mean _the_ npm registry but rather the registry that the package manager is configured to use.

PR #84, opened 2020-12-16
chore: upgrade caniuse-lite db

with-heart:
I'm working on something else in the codebase and noticed that a lot of the snapshots for `run.test` have embedded `Browserslist` errors. This PR upgrades the `caniuse-lite` db and updates the snapshots to keep the snapshot messages directly relevant to the packages.


mitchellhamilton:
Won't this create the same problem when the version updated in this PR is out of date? So isn't leaving it out-of-date kind of good in a way because it won't fail in the future?

with-heart:
Whoops, missed this somehow. Yeap, that makes sense.

PR #85, opened 2020-12-16
feat(cli): add exact pkg/dir name matching for run command

### LINKED ISSUE #82: [`run` command] Add a way to differentiate between packages when one package name is a substring of another
Hi there! Thanks for the great tool. We're using it as part of our tools for managing https://github.com/chakra-ui/chakra-ui to great effect!

One of the problems we've run into is that we have packages where one package name is a substring of another, and as far as I can tell, there isn't a way to force `manypkg` to target the substring name. The packages are named `@chakra-ui/theme` and `@chakra-ui/theme-tools`, so we're stuck without a way to target `@chakra-ui/theme`.

The `README` notes that this can be the case with an example of the `pkg` and `package` directories not being unique, but in this case, the only way around this would be for us to rename one of the packages, since we end up with the same issue in their directory names: `packages/theme`/`packages/theme-tools`.

My first thought was to treat the "partial package name or directory" argument as a regex string, where we'd compare the package name/directory using `match` instead of `includes`. This should allow users to continue using `run` as they did previously, but would also allow us to do something like `manypkg run theme$` to target only `@chakra-ui/theme`.

with-heart:
This PR closes #82 by adding functionality for targeting a project using the exact package or directory name:

`@project/package-a` at `packages/pkg-a`
`@project/package-a-1` at `packages/pkg-a-1`

```bash
yarn manypkg run @project/package-a start
yarn manypkg run packages/pkg-a
```

### DIFF
--- a/.changeset/dirty-phones-tell.md
+++ b/.changeset/dirty-phones-tell.md
@@ -1,0 +1,17 @@
+---
+"@manypkg/cli": minor
+---
+
+Added the ability to use the exact package or directory name to target a package that is a substring of another with for the `run` command:
+
+If packages exist at `packages/pkg-a` and `packages/pkg-a-1`, target `pkg-a` using the exact directory name:
+
+```bash
+yarn manypkg run packages/pkg-a
+```
+
+If packages are named `@project/package-a` and `@project/package-a-1`, target `package-a` using the exact package name:
+
+```bash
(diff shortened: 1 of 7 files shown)


mitchellhamilton (approved):
Thanks!

PR #87, opened 2021-02-01
header structure for rules in the README

acao:
the headers structure for the rules seems irregular and it made it trickier to read, so i set them all to the initial headers structure used for the firs two rules. would you agree that it's harder to see the separation between rules when using h3/h4?

Other ideas to improve readability:
- blockquote the rule description to make it stand out?
- using the collapsing markdown sections GFM supports? doesn't work on npm though?


mitchellhamilton (approved):
Thanks!

PR #108, opened 2021-08-26
Fixed an issue that caused multiple packages to be seleceted

### LINKED ISSUE #107: Error when two packages have same prefix
The issue here is that I have two packages named as `@UI/theme` and `@UI/theme-utils` and when I try to use any scripts with `@UI/theme` the following error is thrown to me:

```sh
☔️ error an identifier must only match a single package but "theme" matches the following packages: 
☔️ error @UI/theme
☔️ error @UI/theme-utils
```

Upon looking at the source code I found the following lines of code that could possibly be causing the issue:
https://github.com/Thinkmill/manypkg/blob/0fecbcb94df338ed1a494c94e30fcacbed2fab5c/packages/cli/src/run.ts#L25-L30

This code snippet lies within the `runCmd` function [here](https://github.com/Thinkmill/manypkg/blob/0fecbcb94df338ed1a494c94e30fcacbed2fab5c/packages/cli/src/run.ts#L25-L30). 

I suppose that `arg[0]` might be the package that is calling the CLI command. So, for example, I am currently calling the CLI command for the `theme` package therefore `arg[0]` should be `@UI/theme`.

With that out of the way now I think `includes` is matching all the packages that have that same name, therefore, causing the issue here. I'll try to make a pull request to fix the issue.

Edit: `args[0]` doesn't equal to `@UI/theme` but just `theme`. I had overlooked some parts of the code

utkarshk384:
Resolves #107


benmccann:
please don't commit a `.DS_Store` file

PR #110, opened 2021-09-09
chore: incorrect type defs

maraisr:
This package imports a type from a file not defined in the package.

https://github.com/Thinkmill/manypkg/blob/24cf58240d8312c050c17708fb0b7e8aad78d21b/packages/get-packages/src/index.ts#L9

May I suggest including `depcheck` to avoid this in the future?

### DIFF
--- a/packages/get-packages/package.json
+++ b/packages/get-packages/package.json
@@ -5,8 +5,8 @@
   "main": "dist/get-packages.cjs.js",
   "dependencies": {
     "@babel/runtime": "^7.5.5",
-    "@manypkg/find-root": "^1.1.0",
     "@changesets/types": "^4.0.1",
+    "@manypkg/find-root": "^1.1.0",
     "fs-extra": "^8.1.0",
     "globby": "^11.0.0",
     "read-yaml-file": "^1.1.0"
--- a/packages/get-packages/package.json
+++ b/packages/get-packages/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@manypkg/find-root": "^1.1.0",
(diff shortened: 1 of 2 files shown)


maraisr:
Friendly bump @Andarist @mitchellhamilton

packages/get-packages/package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@babel/runtime": "^7.5.5",
     "@manypkg/find-root": "^1.1.0",
+    "@changesets/types": "*",
```
Andarist, line 9: That seems to be somewhat dangerous - we can't possibly guarantee that this will work with any potential version of this package. I would much more prefer if this would use the actual latest version with `^`. I know it forces us to bump this range and maintain over time - but at least it's safe.

Andarist:
> May I suggest including depcheck to avoid this in the future?

This package is built using Preconstruct and it already comes with a "variant" of the dep check, unfortunately - it only validates runtime dependencies. I think the type-only deps are kinda ignored by it. Having that improved would be a nice addition to the Preconstruct

maraisr:
In any case, it's a dep that sound be added to the package json. Mind helping me here?

packages/get-packages/package.json
maraisr: Can you suggest a version to use?

mitchellhamilton:
```suggestion
    "@changesets/types": "^4.0.1",
```

maraisr:
@mitchellhamilton @Andarist dang forgot to push up a changesets, like me to PR a subsequent one?

mitchellhamilton:
that would be great

PR #119, opened 2021-10-15
Change external mismatch check & fix to most common

### LINKED ISSUE #117: Configurable fix behaviour
👋 Hey old Thinkmill friends,

We're adopting manypkg in `atlassian-frontend` as a replacement for [Bolt](https://github.com/boltpkg/bolt).
The auto-fix behaviour is awesome, but for dependencies, it fixes to the highest version in the repo.
We'd like it to be configurable to fixing the most used version in the repo.

Because the failure mode of this check usually is because of someone running a unversioned dependency add in a workspace:
```sh
#inside package/packageA
yarn add react
```
This would add the latest version of React, whereas they probably just wanted the version in use in the repo. As usually someone adding a dependency to their package does not mean they also want to upgrade the whole repo.

Happy to contribute this back if this is a change the team is willing to accept.

marcodejongh:
fixes #117

When a developer adds a new dependency, npm/yarn will add the
newest version of the dependency. We want manypkg to fix that newest
version to the most commonly used version in the repo, since it is
unlikely that a developer adding a new dependency to a package also
intends to update the whole repo.

If the dependency range for a dependency all occur once, it will
fix to the highest version.

### DIFF
--- a/packages/cli/src/checks/EXTERNAL_MISMATCH.ts
+++ b/packages/cli/src/checks/EXTERNAL_MISMATCH.ts
@@ -1,6 +1,6 @@
 import {
   makeCheck,
-  getHighestExternalRanges,
+  getMostCommonRangeMap,
   NORMAL_DEPENDENCY_TYPES
 } from "./utils";
 import { Package } from "@manypkg/get-packages";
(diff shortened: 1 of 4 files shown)


marcodejongh:
@mitchellhamilton This should be ready for review. I hadn't thought of the edge-case where all ranges are used the same amount of times, but it was relatively easy to cover off that edge-case, it did lead to me needing a lot more extra tests though.

gautamsi:
how do I use the previous behavior? I was able to fix one dependency and manypkg would do the rest in a large monorepo. is there a flag or argument to `fix` to use previous behavior?

mitchellhamilton:
You can use the [upgrade command](https://github.com/Thinkmill/manypkg/tree/master/packages/cli#manypkg-upgrade-packagename-tag-or-version) for that

PR #122, opened 2021-11-04
fix(pnpm): absolute cwd path matched workspace exclude rules (#121)

### LINKED ISSUE #121: getPackages method not working on cwd absolute path matches the pnpm workspace exclude rules
Now pnpm workspace support workspace exclude rules.

`pnpm-workspace.yaml` case:

```yaml
packages:
  # all packages in subdirs of packages/ and components/
  - 'packages/**'
  # exclude packages that are inside test directories
  - '!**/test/**'
```

so. if cwd absolute path matches the exclude rules, glob will exclude all workspaces.

case (expected results in parentheses) : 

```yaml
  - test
    - packages
      - pkg-1   # (need)
      - pkg-2  # (need)
      - test   # (not need)
    - pnpm-workspace.yaml
```

the parent cwd path has `test` string matches the exclude rule, `getPackages` result is `[]` (empty) .

fz6m:
fix #121

### DIFF
--- a/.changeset/fair-days-rest.md
+++ b/.changeset/fair-days-rest.md
@@ -3,4 +3,4 @@
 "@manypkg/get-packages": patch
 ---
 
-fix absolute cwd path matched pnpm worksapce exclude rules
+Fixed getting correct packages in pnpm workspaces with exclude rules.
--- a/.changeset/fair-days-rest.md
+++ b/.changeset/fair-days-rest.md
@@ -1,0 +1,6 @@
+---
+"@manypkg/cli": patch
+"@manypkg/get-packages": patch
+---
+
+fix absolute cwd path matched pnpm worksapce exclude rules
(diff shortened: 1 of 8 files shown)


.changeset/fair-days-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"@manypkg/cli": patch
+"@manypkg/get-packages": patch
+---
+
+fix absolute cwd path matched pnpm worksapce exclude rules
```
Andarist, line 6:
```suggestion
Fixed getting correct packages in pnpm workspaces with exclude rules.
```

Andarist (approved):
It's a very interesting fix - do you know why this has fixed the mentioned issue? How do those two things relate to each other. I would expect the `absolute: true` to only normalize the final output and not to affect what actually gets matched.

fz6m:
> It's a very interesting fix - do you know why this has fixed the mentioned issue? How do those two things relate to each other. I would expect the `absolute: true` to only normalize the final output and not to affect what actually gets matched.

In fast-glob (globby based on fast-glob).  If `absolute` is specified as `true` :



so first resolve path, then start matching regexp.



like this, start with `!` regexp requires an additional cwd path. but this prefix path is not in the pnpm worksapces rules.

mitchellhamilton (approved):

PR #126, opened 2021-11-12
install commander, update cli to run using it

benkenawell:
The help descriptions could probably use a little tweaking, but this mirrors what the cli was doing before while adding a `help` flag using a robust library (commander) that could do even more to improve the dev experience.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -16,6 +16,7 @@
     "@babel/runtime": "^7.5.5",
     "@manypkg/get-packages": "^1.1.3",
     "chalk": "^2.4.2",
+    "commander": "^8.3.0", 
     "detect-indent": "^6.0.0",
     "find-up": "^4.1.0",
     "fs-extra": "^8.1.0",
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -10,6 +10,7 @@ import { upgradeDependency } from "./upgrade";
 import { npmTagAll } from "./npm-tag";
 import spawn from "spawndamnit";
 import pLimit from "p-limit";
+import { Command } from 'commander';
 
 type RootPackage = Package & {
   packageJson: {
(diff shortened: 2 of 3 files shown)


packages/cli/package.json
```diff
@@ -16,6 +16,7 @@
     "@babel/runtime": "^7.5.5",
     "@manypkg/get-packages": "^1.1.3",
     "chalk": "^2.4.2",
+    "commander": "^8.3.0",
```
Andarist, line 19: The preferred package to handle this sort of stuff would be this: https://www.npmjs.com/package/meow . It's the one used by both Changesets and Preconstruct (and those projects are maintained by roughly the same group of people~)

benkenawell: I see and understand why that makes sense.  I want to argue in favor of commander over meow here, after looking over meow.

1. commander provides a solid api, and docs for that api. https://github.com/commander-cli/commander
    - normally I would prefer meow's method of returning POJOs and arrays, but their documentation is solely lacking relative to commander, and is actually much less flexible than commander's api which is typically the benefit with returning POJOs and the like.
2. commander's help text is automatically generated, it doesn't need maintaining, and won't go out of date.  They're able to do this because you must declare each command and options within their api
3. also due to how commands are called, with commander you can query help for commands individually, which comes for free, whereas you'd have to do that all yourself with meow.
4. Commander allows for a lot of streamlining and simplification of the code you as a maintainer have to write, even after giving you all the above benefits.  Commander takes care of parsing the command line more completely than meow does, and allows for the calling of disparate functions based on commands.

I created PR #128  that accomplishes similar with meow, which you can use for comparison.  As the maintainer, it's, of course, completely up to you which you choose to take (if either!).  I appreciate the work you've done and would love to contribute in the way most helpful to you.

benkenawell:
closing because I stopped receiving feedback from maintainers.

PR #128, opened 2021-12-23
Add help meow

benkenawell:
Adds the [meow package](https://www.npmjs.com/package/meow) to manypkg cli to provide help and version information.  Uses meow because [preconstruct](https://github.com/preconstruct/preconstruct) and [changeset](https://github.com/changesets/changesets) already use it; both packages are maintained by mostly the same group.


benkenawell:
Check out PR #126 for an alternative cli option.

benkenawell:
closing because I never received feedback from the maintainers.

Issue #13, opened 2019-09-19
Overriding external mismatch rule


mitchellhamilton:
Should people be able to override it?

What does configuring it look like?

Noviny:
> Should people be able to override it?

Use-case:

In projects that are large enough, upgrading a package may require changes to a very large surface area. individuals or teams within the project may not have the visibility to bump+test all uses of a package. This creates drag, where no team can bump a package version of anything.

External mismatches can allow you to upgrade across a system incrementally.

> What does configuring it look like?

Less 'what it looks like' but more some design goals:

- Exceptions should be, well, the exception. You should have to actively put yourself in an exceptions state.
- Exceptions should be easy to find + monitor for the repository owner.
- (Most) exceptions should be temporary in nature - with the intent of it being a temporary measure.
- exceptions shouldn't experience a lot of churn (this is an assumption about their usage)
- These assumptions should be enforced/discoverable within the project


Now going to answer some things about the shape:

---
## Formatting

The needed bits of information are:

- external package name
- internal packages where it is an exception
- 'expiry date' for the exception

NB: I don't think any version info needs to be held with the exception - exceptions are whatever, and the latest will be consistent across other packages.

This leads to an exception looking something like:

```json
{
  "packageName": "button",
  "internalPackages": ["pkg-a", "pkg-c"],
  "expiresOn": "YYYY-MM-DD"
}
```

There are ~ two places you can put this: In a special config file just for holding and tracking these, or in the root `package.json` file. I vaguely prefer the ideas of putting these in the root `package.json`, but I'm not 100% on that, as 200 lines of exceptions (unlikely but possible) would be AWFUL.

## manyPkg helpers

(oh no what am I doing creeping this complexity)

### command to add an exception

```
yarn manypkg exception packageName pkg-a,pkg-b
```

### command to list exceptions in order of expiration date

```
yarn manypkg exception list
```

> the wording here is super meh because `exception list` sounds like the name of a thing but 😕


## Thing this design is weird about.

Say I have two packages with a `react-select` exception, and I want them to expire at different dates - does that work? Does it work by just having two entries in the array both specifying `react-select`? What's the CLI process for adding this for two packages at different dates or does it just end up being manual?

Noviny:
(turns out I was carrying a lot of thoughts on this)

mitchellhamilton:
> 'expiry date' for the exception

[...]

Issue #129, opened 2022-01-06
Feature Request: Option Flag to always use Highest Semver Range


fbartho:
Currently, `manypkg check` and `manypkg fix` decide to resolve dependency versions based on the most commonly witnessed range in the repo. -- Only when there are *equal quantities* of each range specifier do we select the highest range:

https://github.com/Thinkmill/manypkg/blob/b8b914441995ad267a490251d443665fb999e32b/packages/cli/src/checks/utils.ts#L152-L158

Would we support a flag `--use-highest-range` on each command to force this algorithm to select the highest range for a given dependency?

Alternative: add a `.manypkg.config.json` this and other flags?

I'm happy to contribute this code change if a maintainer can indicate if they'd accept an appropriate PR implementing this.

jrolfs:
I also prefer this behavior and would be happy to help.

I do worry, though, that y'all are trying to avoid introducing unnecessary configuration... if so is there another recommended way to upgrade a dependency accross an entire monorepo to a specified range?

PR #131, opened 2022-02-03
Allow multiple versions for external mismatch rule

cheesehary:
https://github.com/Thinkmill/manypkg/issues/13#issuecomment-1010619075

We added an extra option to allow multiple versions for the external mismatch check. This allows to upgrade packages incrementally

### DIFF
--- a/packages/cli/src/checks/EXTERNAL_MISMATCH.ts
+++ b/packages/cli/src/checks/EXTERNAL_MISMATCH.ts
@@ -15,7 +15,7 @@ type ErrorType = {
 };
 
 export default makeCheck<ErrorType>({
-  validate: (workspace, allWorkspace) => {
+  validate: (workspace, allWorkspace, rootWorkspace, options) => {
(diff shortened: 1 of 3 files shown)


fbartho:
This looks like it's the building blocks of what I'd need to implement #129 -- did I read this code correctly?

cheesehary:
@fbartho Yes, the current logic is to clamp the version back to the most commonly used one if it breaks the rule. We are thinking to change it to the highest version in the allowed list by default.

Developers might want the highest allowed version most of the time. We can also save some migrations in the incremental upgrade if we install the highest version in the first place.

fbartho:
That sounds great! I certainly vote for this PR *and* for changing the default, or adding an option.

Not sure who is a contributor/maintainer here @cheesehary! Do you think anybody from @thinkmill will approve/merge this?

cheesehary:
@fbartho I'm not sure if anyone is reviewing/will approve this.

PR #143, opened 2022-09-22
Update get-package to @manypkg/get-packages and package-json

barbara79:
The `get-package` library is not maintained anymore see here: https://github.com/changesets/changesets/blob/main/packages/get-workspaces/src/index.ts 
There are some libraries connected to the `get-package` library that are vulnerable.

The same is for the library `package-json`, it has a vulnerable version of the library `got` so it needs to be updated too.

If you could please update the library as I suggested

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -19,7 +19,7 @@
     "detect-indent": "^6.0.0",
     "find-up": "^4.1.0",
     "fs-extra": "^8.1.0",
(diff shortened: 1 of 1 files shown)


benmccann:
It looks like these changes have already landed in `main`: https://github.com/Thinkmill/manypkg/blob/main/packages/cli/package.json

You should be able to close this PR as completed now

PR #151, opened 2022-12-11
Create manypkg/tools and refactor find-root and get-packages

elliot-nelson:
### SUMMARY

Refactor monorepo tools and associated types into a new `core` package, that can be used by both `get-packages` and `find-root`.

_This initial PR is incomplete and I've put it up to gather initial feedback on the approach; if feedback is positive I'll continue to iterate on it. I've currently included a possible API and two tool implementations, Lerna and Rush._

### DETAILS

Today, the process of "adding a tool" to the list of supported tools in manypkg is muddy, because the logic is spread around several packages. Goals:

 - Consolidate all logic related to finding and extracting packages from monorepos into one package.
 - Provide a concrete API for what a "monorepo tool" is, and refactor existing logic to use that API.

### TODO

Next steps / planned approach:

 - [ ] Gather initial feedback
 - [ ] Finish converting remaining tools to this new structure
 - [ ] === (Merge PR 1: Create core package) ===
 - [ ] Modify find-root and get-packages to use `manypkg/core`, and confirm all existing unit tests still pass
 - [ ] Manual validation of functionality on all supported monorepos
 - [ ] === (Merge PR 2: Conversion and publishing of changes) ===
 - [ ] Finally, rewrite all tool-specific unit tests to live on core package (close to the source)
 - [ ] Leave overarching unit tests and tests related to leftover logic in the other packages
 - [ ] === (Merge PR 3: Cleanup / test improvements) ===


packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+    },
+
+    async getPackages(directory: string): Promise<Packages | undefined> {
+        try {
+            const rushText: string = await fs.readFile(path.join(directory, "rush.json"), "utf8");
+
+            // Rush configuration files are full of inline and block-scope comment blocks (JSONC),
+            // so we use a parser that can handle that.
+            const rushJson: RushJson = jju.parse(rushText);
```
Andarist, line 42: TIL about `jju` - nice 👍  I was using recently `golden-fleece` to manipulate JSON5 and I had some formatting issues with it. Definitely gonna give this a try next time.

packages/core/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@manypkg/core",
+  "version": "1.0.0",
+  "license": "MIT",
+  "main": "dist/get-packages.cjs.js",
```
Andarist, line 5: this looks incorrect for this package - you can probably call `preconstruct fix` from the root autoupdate this to the correct value

packages/core/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@manypkg/core",
```
Andarist, line 2: I don't exactly like the `/core` in this name - I don't necessarily have a better name for this just yet though. The boundary line between core/get-packages/find-root is quite blurry to me at the moment. 

Maybe something like `@manypkg/tools` (`@manypkg/project-tools`) would work for this? To pull off smth like this, we'd probably have to remove `expandPackageGlobs` from the public exports of this package though (which doesn't sound to be bad, the `getPackages` methods already call this so it's not necessary to export this for public consumption)

packages/core/src/LernaTool.ts
```diff
@@ -0,0 +1,49 @@
...
+    },
+
+    async getPackages(directory: string): Promise<Packages | undefined> {
+        try {
+            const lernaJson = await fs.readJson(path.join(directory, "lerna.json"));
+            const pkgJson = (await fs.readJson(path.join(directory, "package.json"))) as PackageJSON;
+            const packageGlobs: string[] = lernaJson.packages || ["packages/*"];
+
+            return {
+                tool: LernaTool,
```
Andarist, line 36: returning a tool instead of a tool's name/type here - smart 👍

packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            const directories = rushJson.projects.map(project => path.resolve(directory, project.projectFolder)).sort();
```
Andarist, line 44: It's quite interesting that you `.sort()` this here. Is this strictly necessary?

packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,64 @@
...
+            const packageJsons = await Promise.all(directories.map(dir => fs.readJson(path.join(dir, "package.json"))));
```
Andarist, line 45: So you don't support globs to catch all packages in a given dir? 🤔 Is this list somehow auto-updated by Rush itself or it is always the responsibility of the user to add new packages to the declared list?

packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,64 @@
...
+            const packages = directories.map((dir, index) => {
+              return {
+                dir,
+                packageJson: packageJsons[index] as PackageJSON
+              };
+            });
```
Andarist, line 51: nit: I think that it would be slightly more readable (no index-based zipping) if we'd merge this with the `Promise.all` "loop" a line above

Andarist (commented):
Left a couple of questions/nits/proposals. Overall - I'm excited about this work! :)

packages/core/src/RushTool.ts
elliot-nelson: Rush tends to favor explicit configuration over convention, and doesn't support globs for that reason.

A nice side effect of this approach is that you can guarantee that in order to build and test a new project, the rush.json file must be touched, so your maintainers team can be a codeowners of that file if you want them to approve the initial skeleton PRs for new projects.

packages/core/src/RushTool.ts
elliot-nelson: 🤔 I guess there's a couple options there. I think the answer is _yes_, we do need to sort it, because Rush's own tools (like `rush list`) sorts by package name, and so does the existing `changesets` ascii UI, so users would expect that.

But, do we need to sort it _here_? I think the answer is yes, because I think it's useful if you're implementing a Tool to assume that the list of packages you return will be displayed essentially as-is to users, and not sorted later by get-packages or changesets. This gives the Tool author some flexibility on exactly what "sorting" means -- here it's clearly alphabetical (although perhaps I should make that explicit), whereas if I imagine a later cross-language Tool like a `SwiftProjectYamlTool` or a `AndroidBuildGradleTool`, perhaps an expected ordering of packages is very different.

packages/core/package.json
elliot-nelson: :+1: Also somewhat agnostic on the name, and totally fine if expandPackageGlobs is just some helper method for tool authors and not a public export.

When thinking about the name, though, consider this line of reasoning (which is still only half-baked at the moment)...

 - I'd eventually like `changesets` to be usable even in a monorepo of a different language, like Swift or Kotlin.
 - In these worlds, the file to "read" and "write" might be a totally different format, like a `project.yaml` file or a `gradle.properties` file.
 - In fact the act of "writing" a new version might be totally unrecognizable, perhaps it involves reading or writing several different files for one project.
 - To allow a tool author to contribute such a thing, the surface area of Tool needs to eventually grow to encompass more of what changesets does... the `PackageJSON` object perhaps morphs into a more agnostically named `Project` object, but still requires a name, version, dependency list, etc.; the `Tool` interface grows to include `updateProject(version: SemVer)` or something similar.
 - Some systems allow major, minor, patch, some also semver concepts like -alpha, -prerelease, -snapshot, etc., so a Tool might need to specify what's allowed. In fact, both Swift and Kotlin have complicated relationships with version, because those projects both have "libraries" (released as CocoaPods or Swift Packages, or to Maven) that support full semver spec, but their "applications" (released to App Store or Google Play) are much more explicit and restricted; so there's even a "type" of project we might eventually want to model.

If more and more "functionality" ends up inside Tools, the name `tools` still makes sense so we're probably OK. This also isn't necessarily the right direction (if we decide writing a version and calculating new versions are too similar, then perhaps you add new Tool to manypkg but Changesets still has to add explicit support for that tool in its own list of tool implementations).

elliot-nelson:
Thanks @Andarist, tons of good feedback here!

To summarize:

 - I'll begin chipping away at the various small comments, type adjustments, logic tweaks, etc.
 - I will split apart the replacement of the glob engine, to match changesets, into a separate PR that can go ahead of this one.
 - I will introduce a PackageJSON clone into this PR, but will make sure not to adjust it at all until after changesets switches to it (we can get by on structural typing in meantime).

packages/core/src/LernaTool.ts
```diff
@@ -0,0 +1,49 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+import { PackageJSON } from "@changesets/types";
+import { Tool, ToolType, Package, Packages, MonorepoRoot } from './Tool';
+import { expandPackageGlobs } from "./expandPackageGlobs";
+
+export const LernaTool: Tool = {
```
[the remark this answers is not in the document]
elliot-nelson: Open to playing with satisfies! I wouldn't want to lose the "hard typing" of the object to Tool though, as a tool author, to get that instant red feedback in VSCode if I get out of alignment with the API as I'm writing my method.

packages/core/src/LernaTool.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    async isMonorepoRoot(directory: string): Promise<MonorepoRoot | undefined> {
+        try {
+            const lernaJson = await fs.readJson(path.join(directory, "lerna.json"));
+            if (lernaJson.useWorkspaces !== true) {
+                return {
+                    dir: directory,
+                    tool: LernaTool
+                }
+            }
```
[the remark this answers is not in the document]
elliot-nelson: I'm not that familiar with Lerna, so I'm hoping your own test cases will save me here :)

It's worth noting this logic is pulled right from get-packages, and it's actually saying the opposite of what you might expect -- if lerna.json useWorkspaces IS TRUE, then it is NOT the root of a monorepo.

(If I've misunderstood this, it should come out in the wash when I run your test bed on it.)

elliot-nelson:
Change summary:

 - All tools (Bolt, Lerna, None, Pnpm, Rush, Yarn) are now implemented
 - I've updated the findRoot method and findRoot unit tests to show off the intended new API, which is that _findRoot should return a MonorepoRoot and not a string_. (The idea being it's of limited value to search for a monorepo and then be told "there is one" -- you also need to of what type the monorepo is.) This would be a breaking change for find-root.
 - The `NoneTool` is kind of a special case; it could maybe called `SinglePackageTool`? This MonorepoRoot essentially means there is no monorepo. It's a little odd to search for a monorepo root and get this in return, but this possibility is baked into find-root and get-packages today, so let's keep it working.

packages/get-packages/package.json
```diff
@@ -7,6 +7,7 @@
     "@babel/runtime": "^7.5.5",
     "@changesets/types": "^4.0.1",
     "@manypkg/find-root": "^1.1.0",
+    "@manypkg/core": "^1.0.0",
```
Andarist, line 10: should be sorted:
```suggestion
    "@manypkg/core": "^1.0.0",
    "@manypkg/find-root": "^1.1.0",
```

packages/core/src/LernaTool.ts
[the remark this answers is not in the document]
Andarist: I think it would check all the boxes for us here - you can play around with it [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCoHt0BtkG8BQyRyUEcAJuiFgJ7Jg0AOEAXMgM5hSgDmA3IWLB2AWSrpSjdACVMYABTlgpBGAk02nbiB4BKNgAUo6ALbCIAHgBGmLGRAA+AcWTCxICRCmz0YAMo0IAiKyhCq6ppcvPrINtj2AgC++PgIVJzIADLQIHAY2MgAvHiCDMxsAOR2ULkVyHDsyGkgnAA0gg2BCK6i4pIycvK6JS6kYACuNfRQ4xCCie1Ebn1eA74BQUMjxGOTINOz8-iJHHBgwjDAEI35WCkA9PfIjFhwdHDG4yDkyADuwGAABb1cZqNImF4QSCNODfDgTGAwfDZGp5WwAOhSzQyBhAjBMtyK2zKrGQFUYeJMdQaTXSYEW9XYXR67k83kGwwIoyhewOcyIC2Op3O7Eu1zQtmQj2Q0GMUEa-SgkB+IGASFoQA). It would require us to bump some dev deps and I created a PR for it [here](https://github.com/Thinkmill/manypkg/pull/152)

packages/core/src/LernaTool.ts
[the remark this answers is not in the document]
Andarist: Hm, ye - this seems to be confusing a little bit. `useWorkspaces: true` is required if you want to install your monorepo using the package manager itself (without this Lerna does its own thing). So I guess that we only want to detect a monorepo as a Lerna monorepo when installation is **not** delegated to the package manager. If it's delegated then another tool should just "win".

packages/core/src/RushTool.ts
Andarist: I would probably be very wary of trusting the input to be sorted anyhow. That being said, I don't mind sorting here but I'm also not sure when that "trait" would actually be helpful. This should only matter for display purposes and atm Changesets only display packages by their name (and here we are sorting by the FS path).

packages/core/package.json
Andarist: With all that in mind - `/tools` still makes sense and I would probably go with that. I'm also not that afraid of revisiting the name in the future. It's just that `/core` is such generic that it's hard to tell what it even does.

packages/core/src/BoltTool.ts
```diff
@@ -0,0 +1,85 @@
...
+                throw err;
+            }
+        }
+        return false;
+    },
+
+    async getPackages(directory: string): Promise<Packages> {
+        try {
+            const pkgJson = (await fs.readJson(path.join(directory, "package.json"))) as BoltPackageJSON;
+            const packageGlobs: string[] = pkgJson.bolt!.workspaces!;
```
Andarist, line 47: This could lead to throwing something different than `InvalidMonorepoError` and if we have such a specialized error I would expect that to be thrown in such a case.

packages/core/src/LernaTool.ts
```diff
@@ -0,0 +1,86 @@
...
+        return false;
+    },
+
+    async getPackages(directory: string): Promise<Packages> {
+        try {
+            const lernaJson = await fs.readJson(path.join(directory, "lerna.json"));
+            const pkgJson = (await fs.readJson(path.join(directory, "package.json"))) as PackageJSON;
+            const packageGlobs: string[] = lernaJson.packages || ["packages/*"];
+
+            return {
```
Andarist, line 49: In theory, we should also return `npmClient` here somehow (since Lerna might be configured with different clients). Nobody has ever asked for this though... so I'm totally OK with not doing that

packages/core/src/NoneTool.ts
```diff
@@ -0,0 +1,75 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+import { Tool, ToolType, Package, PackageJSON, Packages, InvalidMonorepoError } from './Tool';
+import { expandPackageGlobs } from "./expandPackageGlobs";
+
+export const NoneTool : Tool = {
```
Andarist, line 7:
```suggestion
export const NoneTool: Tool = {
```

Andarist: In https://github.com/Thinkmill/manypkg/pull/152 I'm bumping Prettier and all - my VS code recognizes it there and uses it to format files on save but perhaps we are missing a precommit hook or something that would help us to keep this clean?

packages/core/src/NoneTool.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    async getPackages(directory: string): Promise<Packages> {
+        try {
+            const pkgJson: PackageJSON = (await fs.readJson(path.join(directory, "package.json")));
+            const pkg: Package = {
+                dir: directory,
+                packageJson: pkgJson
+            };
+
+            return {
```
Andarist, line 42: Related to [the other comment](https://github.com/Thinkmill/manypkg/pull/151#discussion_r1052147465)... `npmClient` would make sense here too 🤔 since a "single"/none project can freely use just any npm client and Changesets might have to use the correct one for publishing and stuff like that.

I'm not sure how this was wired up before and if we had any potential (unreported) issues with this.

Andarist: For the record - since this probably wasn't handled before, we can not worry about this here.

packages/core/src/BoltTool.ts
```diff
@@ -0,0 +1,85 @@
...
+
+            return {
+                tool: BoltTool,
+                packages: await expandPackageGlobs(packageGlobs, directory),
+                root: {
+                    dir: directory,
```
Andarist, line 53: I'm often going back and forth on things like this - but maybe we shouldn't use the abbreviated forms in the public API? I don't feel super strongly about this so I'm probably fine either way.

packages/core/src/PnpmTool.ts
```diff
@@ -0,0 +1,101 @@
...
+            };
+        } catch (err) {
+            if (err.code !== "ENOENT") {
+                throw err;
+            }
+            console.log(err.stack);
+            console.log(new Error().stack);
+            console.log(3);
+            console.log(4);
+            throw err;
```
Andarist, line 73: :P
```suggestion
```

packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,103 @@
...
+            const rushJson: RushJson = jju.parse(rushText);
+
+            const directories = rushJson.projects.map(project => path.resolve(directory, project.projectFolder)).sort();
+            const packages: Package[] = await Promise.all(directories.map(async (dir: string) => {
+                return fs.readJson(path.join(dir, "package.json")).then((packageJson: PackageJSON) => {
+                    return {
+                        dir,
+                        packageJson
+                    };
+                });
```
Andarist, line 58:
```suggestion
                return {
                  dir,
                  packageJson: await fs.readJson(path.join(dir, "package.json")),
                };
```

packages/core/src/Tool.ts
```diff
@@ -0,0 +1,107 @@
...
+export type MonorepoRoot = {
+    dir: string;
+    tool: Tool;
+};
+
+/**
+ * Monorepo tools may throw this error if a caller attempts to get the package
+ * collection from a directory that is not a valid monorepo root.
+ */
+export class InvalidMonorepoError extends Error { }
```
Andarist, line 72: We are constructing the message at all call sites but we could potentially construct it in the `constructor` here.

packages/get-packages/src/index.ts
```diff
@@ -146,20 +71,17 @@ export async function getPackages(dir: string): Promise<Packages> {
...
+
+  return packages;
 }
 
 export function getPackagesSync(dir: string): Packages {
-  const cwd = findRootSync(dir);
-  const pkg = fs.readJsonSync(path.join(cwd, "package.json"));
+  const monorepoRoot: MonorepoRoot = findRootSync(dir);
+  const packages: Packages = monorepoRoot.tool.getPackagesSync(dir);
+  //const pkg = await fs.readJson(path.join(cwd, "package.json"));
```
Andarist, line 82:
```suggestion
```

packages/get-packages/src/index.ts
```diff
@@ -146,20 +71,17 @@ export async function getPackages(dir: string): Promise<Packages> {
...
 
+/*
```
Andarist, line 84: This whole commented out block should be removed

packages/find-root/src/index.ts
```diff
@@ -2,6 +2,8 @@ import findUp, { sync as findUpSync } from "find-up";
 import path from "path";
 import fs from "fs-extra";
 
+import { Tool, ToolType, NoneTool, defaultOrder, supportedTools, MonorepoRoot } from "@manypkg/core";
```
Andarist, line 5: I would probably try to keep the `@manypkg/core` (`/tools`?) as lean as possible. For that reason I would make the `defaultOrder` responsibility of the consumer - and in this case, that would be here, in the `@manypkg/find-root`.

packages/find-root/src/index.ts
```diff
@@ -31,50 +33,60 @@ async function hasWorkspacesConfiguredViaPkgJson(
...
+          monorepoRoot = result;
+        }
 
-async function hasWorkspacesConfiguredViaPnpm(directory: string) {
-  // @ts-ignore
-  let pnpmWorkspacesFileExists = await fs.exists(
-    path.join(directory, "pnpm-workspace.yaml")
+        // the findUp tool expects us to return the current directory or undefined
+        return result ? result.dir : undefined;
+      });
```
Andarist, line 57: nit: could be written like this:
```js
        if (result) {
          monorepoRoot = result;
          // the findUp tool expects us to return the current directory or undefined
          return result.dir;
        }
```

packages/find-root/src/index.ts
```diff
@@ -31,50 +33,60 @@ async function hasWorkspacesConfiguredViaPkgJson(
...
-export async function findRoot(cwd: string): Promise<string> {
+  if (monorepoRoot)
+    return monorepoRoot;
+
+  // No monorepo root was found for any supported tool, so instead we return... something??
+  // I'm not sure what the case for this "default to current root pkg if no monorepo found"
+  // behavior is, it feels like we should return undefined and the calling tool should know
+  // that there is no monorepo.
+
   let firstPkgJsonDirRef: { current: string | undefined } = {
```
Andarist, line 70: I know that you just touched this code but it's written really weirdly. Isn't this whole thing (from this moment to the end of the function) equivalent to smth like?
```ts
findUp(async directory => {
  const exists = await fs.exists(path.join(directory, 'package.json'))
  return exists ? directory : undefined
})
```

packages/find-root/src/index.ts
```diff
@@ -31,50 +33,60 @@ async function hasWorkspacesConfiguredViaPkgJson(
...
   if (firstPkgJsonDirRef.current === undefined) {
     throw new NoPkgJsonFound(cwd);
   }
-  if (dir === undefined) {
-    return firstPkgJsonDirRef.current;
-  }
-  return dir;
+
+  return {
+    tool: NoneTool,
```
Andarist, line 87: You have wanted to add an argument to control what kind of tools can be returned by a particular call. I wonder how this thing fits into that... especially that you didn't include `'none'` in the `defaultOrder` thing.

packages/core/src/LernaTool.ts
elliot-nelson: Very useful feedback here, but let me ask... let's say I'm using Lerna (my monorepo tool), and I'm possibly using one of npm, pnpm, or yarn (my package manager).

What difference would each of these make? Is the choice of tool relevant at all in versioning / changeset consumption / changelog updates, or is it basically at the end of the process the "publish" command only (npm publish vs pnpm publish vs yarn publish).

packages/core/src/NoneTool.ts
elliot-nelson: Most of this is probably my fault -- once I sandpaper some rough edges expect all my code to match whatever standard you already have.

packages/find-root/src/index.ts
elliot-nelson: I can definitely be convinced of this.

Here's my concern -- you've addressed a few corner cases in the existing tools by using "carve outs" in the logic, and I wonder if that is really appropriate. (As an example: LernaTool skips itself if lerna.json specifies useWorkspaces, which lets the logic fall back to YarnTool or PnpmTool.)

It seems like what's more appropriate is to recognize that as the number of Tools grows, you'll end up with Tools that are overlapping or are strict subsets/supersets of each other; the "default order" is an important part of the list of tools, because it is the order that manypkg guarantees will pick the most specific Tool for your repo, falling back to less specific tools.

If we rely on the consumer of the tools to provide a default order, they need to revisit the list every time they update the dependency.

packages/core/src/LernaTool.ts
Andarist: It's sometimes relevant when it comes to publishing - not all package managers handle all of the "available" protocols (although I have a hard time figuring out a scenario in which this would matter for single package repos). The more important difference is that they might handle authentication differently.

But as mentioned, it's actually not a new problem so we can postpone fixing this.

packages/find-root/src/index.ts
```diff
@@ -31,50 +33,60 @@ async function hasWorkspacesConfiguredViaPkgJson(
...
-}
 
-export async function findRoot(cwd: string): Promise<string> {
+  if (monorepoRoot)
+    return monorepoRoot;
+
+  // No monorepo root was found for any supported tool, so instead we return... something??
+  // I'm not sure what the case for this "default to current root pkg if no monorepo found"
+  // behavior is, it feels like we should return undefined and the calling tool should know
+  // that there is no monorepo.
```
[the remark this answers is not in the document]
elliot-nelson: Yeah, despite my misgivings it does seem like passing the buck to the caller sucks.

But, I'm very much not a fan of "NoneTool" 😆 .  Maybe something like "SinglePackageTool" would be better and more clear.

packages/find-root/src/index.ts
elliot-nelson: Yes, it is written really strangely. In the original code before my edits, it was written that way to avoid a second findUp traversal.

I'll fix this up to match your suggested logic, because I think it is _inevitable_ that we need a second findUp if there's no monorepo tool found: especially later, when we allow users to pass in a filtered list of Tools, it just doesn't make sense to be looking for package.json files during the first traversal when the user's chosen tool might not even be "package.json"-based.

packages/find-root/src/index.ts
elliot-nelson: Right -- I implemented logic in `NoneTool` to "find" a package.json file, but that is sort of wrong.

I think because NoneTool (or SinglePackageTool?) is an exception to regular rules, we need to formalize that exception in one of two ways:

 - SinglePackageTool's isMonorepoRoot always returns false no matter what
 - findRoot always excludes SinglePackageTool from its search list if it is included

I think option 1 is probably the best option, because then I can add SinglePackageTool to the end of the default search order, it would never be found as a monorepo root, and then findRoot has the special logic at the end to return it if you are sitting in a regular package folder.

elliot-nelson:
@Andarist You brought up a new concern which I think makes sense to think about in the long term (not, necessarily, in this first stab): package manager variations.

Both LernaTool and RushTool support using NPM, PNPM or Yarn for your "publishing package manager".

Whereas, PnpmTool specifically uses PNPM.

If we imagine a future language-agnostic world, an AndroidTool or GradleTool might provide a _dependency implementation_ (using build.gradle / gradle.properties files), but in order to publish, you can use "gradle publish" or your project might use "maven publish" directly.

If GradleTool did offer a choice between gradle and maven, and Changesets knew how to publish with either, the two publishing tools could require totally different command line options; it means Changesets would really need very specific package manager implementations. I wonder if the concept of package manager should be kept completely contained away from manypkg.

A related concern is that `packages/cli/run.ts` today just assumes you run commands with `yarn`; it might be nice for Tools to specify "commandLineTool", so that manypkg could run multi-package commands using `lerna` in a Lerna repo.

I think the question to answer is: if I have a Lerna repo, and I really want publishing to happen with (npm/pnpm/yarn), how does Changesets know which one to use? I feel like Changesets is the tool that needs to call it, so Changesets needs to decide which one, which means it is Changesets (not manypkg) that will read whatever config file might specify such an option, or invent a configuration option for changesets _itself_ (`changesets.publishingTool`).

Let me know if that hangs together.

packages/core/src/RushTool.ts
elliot-nelson: Good points! I will strip sorting from any Tools that currently do it, and the end consuming UI will do the sorting instead.

elliot-nelson:
@Andarist In a comment somewhere up above, you mentioned:

> I'm unsure whether we should be abbreviating `directory` as `dir` in public APIs

This did get me thinking -- I think that the current API is rather ambiguous, not in terms of dir vs directory, but in terms of whether a given `dir` field is relative or absolute (and what it is relative to). Right now the whole structure returned from `getPackages` is still relative to the original `cwd`, which I think is not ideal.

This is my suggestion on how to fix this:

 - In type `Package`, `dir` -> `relativeDir` -- the name and comment will make it clear that every Package is always expressed as a relative path from the root of the monorepo it lives in
 - In type `Packages`, `root` -> `rootPackage` -- this object always has a `relativeDir` of `.`, because the root package can never have any other relative path
 - Type `Packages` has a new added string property, `rootDir`, which is an _absolute path_ to the root of the monorepo. This cuts off our reliance on the initial `cwd` (search directory).

To give a concrete example, here's today's structure returned vs my proposed change. (In this example, we started the search from the "./packages" folder.)

### Old:

```javascript
{
  tool: "pnpm",
  packages: [
    { dir: "packages/a", packageJson: { name: "a", /* ... */ } },
    { dir: "packages/b", packageJson: { name: "b", /* ... */ } }
  ],
  root: { dir: "..", packageJson: { name: "myrepo", /* ... */ } }
}
```

### New:

```javascript
{
  tool: { type: "pnpm", /* ... */ },
  packages: [
    { relativeDir: "packages/a", packageJson: { name: "a", /* ... */ } },
    { relativeDir: "packages/b", packageJson: { name: "b", /* ... */ } }
  ],
  rootPackage: { relativeDir: ".", packageJson: { name: "myrepo", /* ... */ } },
  rootDir: "/Users/janjansen/dev/myrepo"
}
```


Let me know what you think.

packages/find-root/src/index.ts
Andarist: We could do this:
1. move `defaultOrder` to `find-root`
2. require the user to opt into keyed tools with smth like `{ rush: true }`. 

This way we would still be in charge of the order while also allowing the user to specify what they allow to be detected.

packages/find-root/src/index.ts
Andarist: Sounds good when skimming over this. Could you implement this change (if you didn't do it already)?

Andarist:
> A related concern is that packages/cli/run.ts today just assumes you run commands with yarn;

Oh, I didn't realize this. It's definitely "bad" and we need to change this somehow. I never used `manypkg` to run scripts/commands, this was implemented in https://github.com/Thinkmill/manypkg/pull/42 but in the today's landscape... I question it's usefulness. We don't want to compete with Turborep, Nx and others here. We can keep it around as a lightweight solution for this problem but dunno, I would probably be fine with removing it.

> I feel like Changesets is the tool that needs to call it, so Changesets needs to decide which one, which means it is Changesets (not manypkg) that will read whatever config file might specify such an option, or invent a configuration option for changesets itself (changesets.publishingTool).

I agree on principle that this is way more relevant for Changesets than it is for manypkg. OTOH, having yet another configuration/auto-selection thing in there (that should also be somewhat tightly-coupled with a Tool) doesn't sound like something fun to use. Either way - it's a problem for another day.

---

Re `relativeDir` and so on: 👍

Re abbreviations - to sum it up, you prefer the abbreviated `dir` over `directory`, right?

>   tool: { type: "pnpm", /* ... */ },

I think that you didn't describe this change but it's in the proposed new format.

>   rootPackage: { relativeDir: ".", packageJson: { name: "myrepo", /* ... */ } },
>  rootDir: "/Users/janjansen/dev/myrepo"

Alternative could look something like:
```
root: {
    dir: "/Users/janjansen/dev/myrepo",
    packageJson: { name: "myrepo", /* ... */ } // optional
}
```

But I guess that this would make it *more* annoying to work with as `root` wouldn't be of a `Package` type. Your proposal sounds better. I think I might have asked this somewhere... or maybe I just had this thought in the couple of last days. Do we even need to have a special `rootPackage`? Isn't this just a package? I can imagine that it might be a special package but I don't see any actual argument for it right now (one could be found further down the road so it still might be worth it to keep it separately). I can easily miss something obvious here though.

elliot-nelson:
@Andarist

> Re abbreviations - to sum it up, you prefer the abbreviated dir over directory, right?

Yeah, I don't see a need for the extra chars; I think clarifying `rootDir` vs `relativeDir` is more useful than the long name. I'd also accept `rootFolder` and `relativeFolder` as halfway in-between (shorter than Directory but not abbreviated 😆).

> Do we even need to have a special rootPackage? Isn't this just a package?

If we take up the proposal above, and eliminated rootPackage, you could _find_ the root package with `.packages.find(p => p.relativeDir === ".")`.

However, I think that actually in most repos that are actually monorepos, if the root package exists, it is a "dev dependency dumpster" -- it just exists to throw compilers and linters and CLI tools and libraries at, it's never intended to be published and therefore isn't actually part of your dependency graph.

Changeset's `get-dependents-graph` does include the root package, _but_ I don't think that code is actually claiming "all the projects and the root project go in the graph" -- I think it's more like "well, we might ONLY have a root project, which is not a monorepo, so all the changesets will be for that one; or we might have a bunch of packages, in which case probably all the changesets are for those".

Andarist:
> Yeah, I don't see a need for the extra chars; I think clarifying rootDir vs relativeDir is more useful than the long name. I'd also accept rootFolder and relativeFolder as halfway in-between (shorter than Directory but not abbreviated 😆)

Let's roll with `rootDir` and `relativeDir` then 👍 

> However, I think separating rootPackage has value in repos that are actually monorepos, if the root package exists, it is a "dev dependency dumpster" -- it just exists to throw compilers and linters and CLI tools and libraries at, it's never intended to be published and therefore isn't actually part of your dependency graph.

Totally agree with that. Some tools are even **requiring** `private: true` in that root `package.json`. That being said - that `package.json` can refer to some local workspaces/packages so it should be the part of the dependency graph (as we might need to update the contained ranges there when versioning)

packages/core/src/Tool.ts
elliot-nelson: If you'd really like to see this, I can push the template into the Error.

Right now I was relying more on convention because I expect the Tools to get more complex / detailed errors in some cases, e.g. `Directory blah is not a valid Blah monorepo: missing bolt property in package.json`, etc. So being able to pass in any string lets a Tool author customize the messaging, while still letting upstream code detect the entire class of error.

Andarist: Let's stick to a simple string

packages/core/src/RushTool.ts
```diff
@@ -0,0 +1,115 @@
...
+    const rootDir = path.resolve(directory);
+
+    try {
+      const rushText: string = await fs.readFile(path.join(directory, 'rush.json'), 'utf8');
+
+      // Rush configuration files are full of inline and block-scope comment blocks (JSONC),
+      // so we use a parser that can handle that.
+      const rushJson: RushJson = jju.parse(rushText);
+
+      console.log(rushJson, 'rush.json');
```
Andarist, line 53:
```suggestion
```

packages/core/src/BoltTool.ts
```diff
@@ -0,0 +1,97 @@
...
+        } catch (err: any) {
+            if (err.code !== "ENOENT") {
+                throw err;
+            }
+        }
+        return false;
+    },
+
+    async getPackages(directory: string): Promise<Packages> {
+        const rootDir = path.resolve(directory);
```
Andarist, line 45: If we are calling `path.resolve` here then I feel like probably we should use `rootDir` over `directory` throughout this function - unless there is a strong reason to use `directory` at any given place.

This comment applies to all `getPackages(Sync)` - since they all follow the same structure

packages/core/src/SinglePackageTool.ts
```diff
@@ -0,0 +1,67 @@
+import path from 'path';
+import fs from 'fs-extra';
+
+import { Tool, ToolType, Package, PackageJSON, Packages, InvalidMonorepoError } from './Tool';
+import { expandPackageGlobs } from "./expandPackageGlobs";
+
+export const SinglePackageTool: Tool = {
+    type: 'package',
```
Andarist, line 8: This probably should match the value declared in this union:
https://github.com/Thinkmill/manypkg/blob/3191db736c686119f3c3da9971872792bf7b8775/packages/core/src/Tool.ts#L4

packages/core/src/Tool.ts
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * The pre-loaded package json structure.
+     */
+    packageJson: PackageJSON;
+
+    /**
+     * Path to the directory containing this package, relative to the monorepo root (for
+     * a "root package", this is the string ".").
+     */
+    relativeDir: string;
```
Andarist, line 51: Last minute thought - `relativeDir` is useful for display purposes (for error messages and stuff like that) but for everything else, it's less useful as it's not sufficient for manipulating the package (you always have to `join` it with the `rootDir`). Perhaps it would be better to have an absolute path here? 🤔

packages/cli/src/__snapshots__/run.test.ts.snap
```diff
@@ -51,17 +51,33 @@ exports[`Run command should execute "package-three start" and exit with 1: stder
 "
 `;
 
+exports[`Run command should execute "package-three start" and exit with 1: stderr 2`] = `
```
Andarist, line 54: Hmmm, the changes in those snapshots are a little bit surprising as the CLI tests are untouched. Do you happen to know where they are coming from?

packages/core/src/LernaTool.ts
```diff
@@ -0,0 +1,92 @@
...
+            }
+        } catch (err: any) {
+            if (err.code !== "ENOENT") {
+                throw err;
+            }
+        }
+        return false;
+    },
+
+    async getPackages(directory: string): Promise<Packages> {
```
Andarist, line 43: an interesting fact about this one - this can be successfully called even when `isMonorepoRoot` returns `false` (since `getPackages` doesn't validate again the `useWorkspaces` flag). We won't be doing that since we'll always check `isMonorepoRoot` beforehand but I wonder if this is OK in terms of the design.

packages/core/src/YarnTool.ts
```diff
@@ -0,0 +1,93 @@
...
+        }
+        return false;
+    },
+
+    async getPackages(directory: string): Promise<Packages> {
+        const rootDir = path.resolve(directory);
+
+        try {
+            const pkgJson = (await fs.readJson(path.join(directory, "package.json"))) as YarnPackageJSON;
+            const packageGlobs: string[] = Array.isArray(pkgJson.workspaces) ? pkgJson.workspaces : pkgJson.workspaces!.packages;
```
Andarist, line 51: Similar doubt here as with Bolt before, this could make this function throw a `TypeError` instead of `InvalidMonorepoError`. Should we use the same strategy as with Bolt?
https://github.com/Thinkmill/manypkg/blob/3191db736c686119f3c3da9971872792bf7b8775/packages/core/src/BoltTool.ts#L49-L51

packages/core/src/Tool.ts
elliot-nelson: This is a tough call because no matter what choice you make, you are visiting pain upon the opposing side -- doing `path.relative()` to get a display path feel onerous, but so does doing `path.join()` to get an absolute path.

In Rush, the (equivalent) of this object solves this by providing both -- a `projectFolder` for the absolute path and the `relativeProjectFolder` for the relative path.

Maybe it makes sense to do the same here, bake in both paths in the returned object and then you don't have to choose?

elliot-nelson:
Latest batch of updates:

 - The `core` package is now named `tools`
 - Fixed up all the outstanding yarn type errors and warnings in the consuming packages
 - Decided on `if (err && (err as any).code == "ENOENT")` as the boilerplate for enoent errors -- see what you think.

Outstanding conversations/open questions as of latest commit:

 - Should `Package` provide relative dir, absolute dir, or both?
 - todo: Move defaultOrder from tools to find-root
 - Should "defaultOrder" and "supportedTools" be exported as constants (by that I mean _look_ like constants, e.g., `SUPPORTED_TOOLS`). Actually, maybe supported tools is just a bad name; it's more like "tools by name".

packages/cli/src/__snapshots__/run.test.ts.snap
elliot-nelson: This should be fixed in the latest push.

Note that I was having a lot of trouble running these tests locally -- it turns out the bootup time of loading the babel "inline" transpiler to run the cli can take exactly on the edge of 5 seconds, which is the default unit test timeout. (I bumped the timeout to 10 seconds for the cli test and they are a lot more stable on my MacBook now.)

packages/core/src/Tool.ts
elliot-nelson: I've thought about this some more, and while I finish up remaining references in manypkg, I'm realizing that the two are not as equivalent as I claimed.

The difference is: if we go with absolute everything, you _can_ deduce the relative path; but the opposite is not true (given only a Package object).

This has pretty big implications for API surface area. For example, if you write a method that takes a single `Package` object, that method can read/update the json, write out updates, search for sibling packages, etc; none of that is possible if `Package` contains only a relative directory, because you don't actually know where it is.

I think, then, the right call is to go back to `dir` for all packages. (But we can still have `rootDir` at the Packages level.)

I'm a bit bummed about the `Packages` name when I see it used in other modules because it doesn't make a lot of sense. `PackageCollection` would be better, but _really_, it's a `Monorepo` -- a root dir, a collection of packages, an optional root package, a tool implementation, that's a monorepo. The fact that it _can_ be used to represent a single package maybe shouldn't dissuade us from just calling it that.

packages/cli/src/run.test.ts
```diff
@@ -34,6 +34,7 @@ describe("Run command", () => {
       expect(code).toBe(exitCode);
       expect(stripAnsi(stdout.toString())).toMatchSnapshot("stdout");
       expect(stripAnsi(stderr.toString())).toMatchSnapshot("stderr");
-    }
+    },
+    10_000 /* 10 seconds */
```
Andarist, line 38: that seems excessive 😅 perhaps it was just committed after debugging something? is there any reason why those tests would be taking so much time?

```suggestion
    }
```

packages/cli/src/index.ts
```diff
@@ -134,22 +134,24 @@ async function execCmd(args: string[]) {
...
     packages.map((x) => [x.packageJson.name, x])
   );
-  packagesByName.set(root.packageJson.name, root);
+  if (rootPackage) {
+    packagesByName.set(rootPackage.packageJson.name, rootPackage);
+  }
   let { hasErrored, requiresInstall } = runChecks(
     packagesByName,
-    root,
+    rootPackage!,
```
Andarist, line 154: This non-null assertion doesn't look right. I think I managed to juggle things around a little bit to fix it here:
https://github.com/elliot-nelson/manypkg/pull/3

packages/find-root/package.json
```diff
@@ -6,6 +6,7 @@
   "module": "dist/find-root.esm.js",
   "dependencies": {
     "@babel/runtime": "^7.20.6",
+    "@manypkg/tools": "1.0.0",
```
Andarist, line 9:
```suggestion
    "@manypkg/tools": "^1.0.0",
```

packages/get-packages/src/index.ts
```diff
@@ -6,21 +6,19 @@ import fs from "fs-extra";
 import path from "path";
 import globby, { sync as globbySync } from "globby";
 import readYamlFile, { sync as readYamlFileSync } from "read-yaml-file";
-import { PackageJSON } from "@changesets/types";
 import { findRoot, findRootSync } from "@manypkg/find-root";
-
-const isNoEntryError = (err: unknown): boolean =>
-  !!err && typeof err === "object" && "code" in err && err.code === "ENOENT";
-
-export type Tool = "yarn" | "bolt" | "pnpm" | "lerna" | "root";
```
Andarist, line 15: note that `'root'` -> `'singlePackage'` can potentially be a breaking change if `.type` is returned anyhow to the caller

packages/get-packages/src/index.ts
```diff
@@ -6,21 +6,19 @@ import fs from "fs-extra";
...
+  ToolType,
+  Package,
+  Packages,
+  MonorepoRoot,
+  PackageJSON,
+  supportedTools,
+  defaultOrder,
+} from "@manypkg/tools";
+
+export { Package, Packages } from "@manypkg/tools";
```
Andarist, line 21: Since we are using `isolatedModules` this is required here:
```suggestion
export type { Package, Packages } from "@manypkg/tools";
```

packages/get-packages/src/index.ts
```diff
@@ -6,21 +6,19 @@ import fs from "fs-extra";
 import path from "path";
 import globby, { sync as globbySync } from "globby";
 import readYamlFile, { sync as readYamlFileSync } from "read-yaml-file";
```
Andarist, line 8: Those are no longer used
```suggestion
```

packages/get-packages/src/index.ts
```diff
@@ -6,21 +6,19 @@ import fs from "fs-extra";
...
-
-export type Packages = {
-  tool: Tool;
-  packages: Package[];
-  root: Package;
-};
+import {
+  Tool,
+  ToolType,
+  Package,
```
Andarist, line 13: Those aren't used as well
```suggestion
```

packages/get-packages/src/index.ts
```diff
@@ -6,21 +6,19 @@ import fs from "fs-extra";
...
+  Packages,
+  MonorepoRoot,
+  PackageJSON,
+  supportedTools,
+  defaultOrder,
```
Andarist, line 18: Those aren't used as well
```suggestion
```

Andarist:
> Should Package provide relative dir, absolute dir, or both?

I don't have strong opinions about this. Relative dir is better for printing error messages, and absolute dir it better for writing back to those files/packages. Each has its own cons/pros and one can't be derived from the other one alone - we need the `rootDir` to derive them. So perhaps having both provide the best DX for our implementation etc at the expense of a harmless redundancy? 

I will be fine either way though - pick our poison :p

packages/get-packages/src/index.ts
elliot-nelson: This is going to be a breaking change anyway because packages that used manypkg before will need to adjust a check like `monorepo.tool === "root"` to `monorepo.tool.type === "root"` -- but I agree that changing the property name (and not the values) does make this a much easier transition.

I've renamed the tool back to "root" / "RootTool", to ease initial transition to the new structure.

packages/cli/src/run.test.ts
elliot-nelson: On my mac, just running `node packages/cli/bin.js` takes almost 5 seconds just to print out an error, so the default timeout of 5 seconds is way too short to reliably run the unit tests. Not sure if this is misconfiguration and/or node versions or whatever, but starting up preconstruct takes a very long time.

packages/get-packages/src/index.ts
Andarist: Ah, ye - good point about the changed structure. It's not like we have a lot of API consumers though, so I don't think that we have to be too cautious when it comes to breaking changes. If we believe that the new name was better, let's keep it. It's better to clean up what we can now, in a single sweep - rather than to drag it through multiple major versions or keep the confusing name forever.

packages/cli/src/run.test.ts
Andarist: Hmm, perhaps `it.each` is counted as a single test and the timeout is shared across all cases. Maybe using `it.concurrent.each` would fix this? 🤔 I can dig into this later to confirm how the timeout is calculated

elliot-nelson: I don't think the issue is with the tests -- what I'm saying is, let's say I just sit at a regular Terminal, booting up preconstruct takes an extra 5 real seconds compared to running a pre-compiled (ts -> js) file. It takes about 5 seconds to show the first line of console output.

So, there's nothing "wrong" with Jest -- running the tool takes ~5 seconds, so every individual test has a 50% chance of failing because it's right on the edge of the timeout.

elliot-nelson:
Latest recap:

1. I have added a .changeset entry for this PR, check it out and let me know if you think it is sufficient.

2. I am suggesting a slightly different change for the try/catch guards -- I think although it's maybe another or two of code, this pattern is easier to read and easier to extend later (if, for example, you want to handle multiple different error codes). Let me know what you think:

```javascript
  isMonorepoRootSync(directory: string): boolean {
    try {
      const lernaJson = fs.readJsonSync(
        path.join(directory, "lerna.json")
      ) as LernaJson;
      if (lernaJson.useWorkspaces !== true) {
        return true;
      }
    } catch (err) {
      if (err && (err as any).code === "ENOENT") {
        return false;
      }
      throw err;
    }
    return false;
  },
```

I think that's all that is left outstanding!

packages/tools/src/LernaTool.ts
```diff
@@ -0,0 +1,118 @@
...
+  type: "lerna",
+
+  async isMonorepoRoot(directory: string): Promise<boolean> {
+    try {
+      const lernaJson = (await fs.readJson(
+        path.join(directory, "lerna.json")
+      )) as LernaJson;
+      if (lernaJson.useWorkspaces !== true) {
+        return true;
+      }
```
Andarist, line 32:
```suggestion
      // we only detect monorepo as a Lerna one when it's using Lerna
      // and when package installation is handled by it and *not* delegated to the workspace-aware package manager
      if (lernaJson.useWorkspaces !== true) {
        return true;
      }
```

packages/cli/src/checks/__tests__/INCORRECT_REPOSITORY_FIELD.ts
```diff
@@ -169,9 +169,9 @@ describe("incorrect repository field", () => {
       (rootWorkspace.packageJson as any).repository =
         "https://dev.azure.com/Thinkmill/monorepos/_git/manypkg";
 
-      rootWorkspace.dir = __dirname;
+      rootWorkspace.relativeDir = ".";
```
Andarist, line 172: it seemed a little bit weird that we have to mutate this stuff here, it's outside of the scope of this PR so doesn't have to be fixed here - I prepared a small PR targeting yours though: https://github.com/elliot-nelson/manypkg/pull/5

packages/cli/src/checks/utils.ts
```diff
@@ -52,7 +50,7 @@ type AllCheck<ErrorType> = {
   validate: (
     workspace: Package,
     allWorkspaces: Map<string, Package>,
```
Andarist, line 52: this wasn't introduced by this PR so we can clean this up later - but I think we should call those either workspaces or packages, right now the codebase is using two different names for the same thing

packages/cli/src/index.ts
```diff
@@ -1,3 +1,5 @@
+// @flow
```
Andarist, line 1:
```suggestion
```

packages/cli/src/index.ts
```diff
@@ -96,13 +89,13 @@ let runChecks = (
...
 async function execCmd(args: string[]) {
-  let { packages } = await getPackages(process.cwd());
+  let { packages, rootDir } = await getPackages(process.cwd());
   let highestExitCode = 0;
   await Promise.all(
     packages.map((pkg) => {
       return execLimit(async () => {
         let { code } = await spawn(args[0], args.slice(1), {
-          cwd: pkg.dir,
+          cwd: path.join(rootDir, pkg.relativeDir),
```
Andarist, line 98: isn't this equivalent to this right now?:
```suggestion
          cwd: pkg.dir,
```

packages/cli/src/run.test.ts
Andarist: hm, ok - let's roll with that. We probably could speed this up significantly by not going through `spawn` and co. We could invoke "command functions" directly in tests

packages/cli/src/run.ts
```diff
@@ -5,18 +5,15 @@ import * as logger from "./logger";
...
-      pkg.packageJson.name === args[0] ||
-      path.relative(root.dir, pkg.dir) === args[0]
-    );
+    return pkg.packageJson.name === args[0] || pkg.relativeDir === args[0];
   });
 
   if (exactMatchingPackage) {
     const { code } = await spawn("yarn", args.slice(1), {
-      cwd: exactMatchingPackage.dir,
+      cwd: path.join(rootDir, exactMatchingPackage.relativeDir),
```
Andarist, line 16: most likely this can be simplified here too
```suggestion
      cwd: exactMatchingPackage.dir,
```

packages/cli/src/run.ts
```diff
@@ -43,7 +40,7 @@ export async function runCmd(args: string[], cwd: string) {
     throw new ExitError(1);
   } else {
     const { code } = await spawn("yarn", args.slice(1), {
-      cwd: matchingPackages[0].dir,
+      cwd: path.join(rootDir, matchingPackages[0].relativeDir),
```
Andarist, line 43: most likely this can be simplified here too
```suggestion
      cwd: matchingPackages[0].dir,
```

packages/tools/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * A default ordering for monorepo tool checks.
+ *
+ * This ordering is designed to check the most typical package.json-based
+ * monorepo implementations first, with tools based on custom file scchemas
+ * checked last.
+ */
+const defaultOrder: ToolType[] = [
```
Andarist, line 28: I think the default order is mainly related to finding the root and stuff like that so it makes sense to move it to `/find-root`. I wanted to know how that would feel like so I prepared a PR here: https://github.com/elliot-nelson/manypkg/pull/6

It's definitely not a hill that I want to die on but it would be kinda neat if all exports from `/tools` would actually be Tools. This way you can utilize `import *` better etc.

Andarist: Let me know what do you think

packages/find-root/src/index.ts
```diff
@@ -98,50 +118,51 @@ function hasWorkspacesConfiguredViaPkgJsonSync(
...
-    return directory;
-  }
-}
 
-export function findRootSync(cwd: string) {
+  if (monorepoRoot) return monorepoRoot;
+
+  // If there is no monorepo root, but we can find a single package json file, we will
+  // return a "RootTool" repo, which is the special case where we just have a root package
+  // with no monorepo implementation (i.e.: a normal package folder).
```
Andarist, line 145: how does it play with the root tool being in the `defaultOrder` **and** with the fact that it's not even the last tool there? I didn't have time to analyze the situation to conclude why this is the case - but it definitely caught my eye and interest

Andarist:
> I am suggesting a slightly different change for the try/catch guards -- I think although it's maybe another or two of code, this pattern is easier to read and easier to extend later (if, for example, you want to handle multiple different error codes). Let me know what you think:

Could you present this as a diff? I'm not completely sure what are we comparing here.

.changeset/curly-days-hang.md
```diff
@@ -0,0 +1,17 @@
...
+
+Release a new API for monorepo tools
+
+ - The `find-root` package now returns a `MonorepoRoot` interface, instead of a string. This
+   interface provides both a `rootDir` for the discovered monorepo, and a `tool`, representing
+   the monorepo tool that was discovered.
+ - The `get-packages` package now returns a slightly different structure. The old `tool` string
+   is replaced with the `tool` object above, and each `Package` now contains both the absolute
+   directory and a relative directory path. The `root` package has been renamed `rootPackage`
+   and is now optional, to support monorepos that do not contain a root package.
```
Andarist, line 16: I'd probably split this information into separate changesets - we don't have to maintain 1 to 1 relationship between a PR and a changeset. Each package could be documented from its own PoV (sometimes at the expense of duplicating some information). Sometimes it definitely makes sense to reuse the message completely, but I feel like perhaps this is not the case - especially since we, currently, maintain a CHANGELOG per package (and not an "aggregated" CHANGELOG that documents the whole monorepo)

elliot-nelson:
> > I am suggesting a slightly different change for the try/catch guards -- I think although it's maybe another or two of code, this pattern is easier to read and easier to extend later (if, for example, you want to handle multiple different error codes). Let me know what you think:
> 
> Could you present this as a diff? I'm not completely sure what are we comparing here.

@Andarist Ah, that was silly, I forgot to push the commit!

This is what I was referring to: https://github.com/Thinkmill/manypkg/pull/151/commits/b8ed304b861233011e4ce527d3ed67e6d055e39d#diff-b9a9e78d290f3682854230f477271a105f907e1cea680ac208672a42b6358fcdR50

```
    } catch (err) {
      if (err && (err as any).code === "ENOENT") {
        return false;
      }
      throw err;
    }
    return false;
```

My (lukewarm) stance is that it's more useful to have a bog-standard, easy to read (and copy) clause for handling error codes, than it is to minimize the number of total exit points. So in my proposal, you're duplicating the `return false` exits (you could exit due to ENOENT, or due to falling out of the other checks above), but, the err checks themselves are "normal".

packages/find-root/src/index.ts
elliot-nelson: Fixed -- this was just me mis-ordering them, I think we want RootTool to always come last (technically not required, but makes the most sense to the reader I think).

packages/tools/src/index.ts
elliot-nelson: Love it, merged into the PR!

packages/tools/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { BoltTool } from "./BoltTool";
+import { LernaTool } from "./LernaTool";
+import { PnpmTool } from "./PnpmTool";
+import { RootTool } from "./RootTool";
+import { RushTool } from "./RushTool";
+import { YarnTool } from "./YarnTool";
+
+export * from "./Tool";
+export { BoltTool, LernaTool, PnpmTool, RootTool, RushTool, YarnTool };
```
Andarist, line 9: nit:
```suggestion
export { BoltTool } from "./BoltTool";
export { LernaTool } from "./LernaTool";
export { PnpmTool } from "./PnpmTool";
export { RootTool } from "./RootTool";
export { RushTool } from "./RushTool";
export { YarnTool } from "./YarnTool";

export * from "./Tool";
```

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * A package.json access type.
+ */
+export type PackageAccessType = "public" | "restricted";
+
+/**
+ * An in-memory representation of a package.json file.
+ */
+export type PackageJSON = {
```
Andarist, line 14: nit:
```suggestion
export interface PackageJSON {
```

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
...
+  manypkg?: {
+    [key: string]: string;
+  };
+};
+
+/**
+ * An individual package json structure, along with the directory it lives in,
+ * relative to the root of the current monorepo.
+ */
+export type Package = {
```
Andarist, line 44: nit
```suggestion
export interface Package {
```

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
...
+   * root (for a "root package", this is the string ".").
+   */
+  relativeDir: string;
+};
+
+/**
+ * A collection of packages, along with the monorepo tool used to load them,
+ * and (if supported by the tool) the associated "root" package.
+ */
+export type Packages = {
```
Andarist, line 66: nit
```suggestion
export interface Packages {
```

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * An object representing the root of a specific monorepo, with the root
+ * directory and associated monorepo tool.
+ *
+ * Note that this type is currently not used by Tool definitions directly,
+ * but it is the suggested way to pass around a reference to a monorepo root
+ * directory and associated tool.
+ */
+export type MonorepoRoot = {
```
Andarist, line 96: nit
```suggestion
export interface MonorepoRoot {
```

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
...
+   * The underlying tool implementation for this monorepo.
+   */
+  tool: Tool;
+};
+
+/**
+ * Monorepo tools may throw this error if a caller attempts to get the package
+ * collection from a directory that is not a valid monorepo root.
+ */
+export class InvalidMonorepoError extends Error {}
```
Andarist, line 112: Two things here.

We are still using the old Babel config that transpiles a lot. Subclassing builtins has always been a problem when transpiled to ES5. Since this whole PR is a breaking change anyway - it's a good time to bump our syntax level, so we should tweak the config here:
https://github.com/Thinkmill/manypkg/blob/45b396d8c940db8d7ad0b9d6f9dd8dc695a15312/babel.config.js#L2

This can be configured like [here](https://github.com/changesets/changesets/blob/8447c677a37d1d23858059e0ce2b5cd98f4c24ae/babel.config.js#L3-L8) but we can already use `node: 14`.

The same information should be added to `package.json#engines`

A second thing is the question - should we export this publicly? I would probably prefer to avoid this because:
- this is a runtime export - so if we export this then we no longer get this: `import * as tools from '@manypkg/tools'`
- in more complex scenarios, node_modules layout etc relying on `instanceof` *externally* is a footgun anyway - so perhaps we shouldn't make it easy for people.

packages/tools/src/Tool.ts
```diff
@@ -0,0 +1,147 @@
+/**
+ * A unique string identifier for each type of supported monorepo tool.
+ */
+export type ToolType = "bolt" | "lerna" | "pnpm" | "root" | "rush" | "yarn";
```
Andarist, line 4: An additional question related your doubts regarding `.type` - if we don't want people to rely on this etc, should we even export this union?

packages/find-root/src/index.ts
Andarist: So the `RootTool` is still in the `defaultOrder` etc but it actually can't ever be matched through that cause it just returns `false` from `isMonorepoRoot`. Is it even worth keeping it there then?

.changeset/frozen-yogurt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@manypkg/tools": minor
+---
+
+Create a new API for monorepo tools
+
+- Introduces a new `Tool` API that provides key functions related to a specific implementation of a monorepo, like `isMonorepoRoot` and `getPackages`. Existing tool implementations in manypkg have been converted to use this new interface.
```
Andarist, line 7:
```suggestion
Introduces a new `Tool` API that provides key functions related to a specific implementation of a monorepo, like `isMonorepoRoot` and `getPackages`. Existing tool implementations in manypkg have been converted to use this new interface.
```

packages/tools/src/BoltTool.ts
```diff
@@ -0,0 +1,125 @@
+import path from "path";
```
Andarist, line 1: I think that perhaps it's time to drop this tool - the project has stalled, isn't actively maintained and has a minimal amount of downloads

packages/tools/src/Tool.ts
elliot-nelson: I think not -- I will remove altogether.

packages/tools/src/Tool.ts
elliot-nelson: I'm not dead-set on exporting it publicly, but I'll push back a little:

> this is a runtime export - so if we export this then we no longer get this: import * as tools from '@manypkg/tools'

I'm not sure what you mean here -- if you mean that `tools` is a collection of Tools, that is already not true, because `import *` also includes all our tool utilities like `Tool`, `PackageJSON`, `Packages`, etc. I think `@manypkg/tools` is a toolbox, and contains a collection of pre-baked Tools and all the ingredients you need to roll your own, and this Error is one of those ingredients.

> in more complex scenarios, node_modules layout etc relying on instanceof externally is a footgun anyway - so perhaps we shouldn't make it easy for people.

I suppose this is true although I do `instanceof` on Errors exported by e.g. AWS SDK without issue. If we wanted to truly fix this then I'd recommend adding a `static isInvalidMonorepoError(error: unknown)` boolean onto the Error, and adding a property `code: "InvalidMonorepoError"` to the class to perform the check. In this case I'd still argue it's useful to export because now you can use `if (InvalidMonorepoError.isInvalidMonorepoError(error))` in a try catch block.

Andarist: > I'm not sure what you mean here -- if you mean that tools is a collection of Tools, that is already not true, because import * also includes all our tool utilities like Tool, PackageJSON, Packages, etc.

All of those are just types - not runtime values. 
```ts
Object.keys(require('@manypkg/tools')) // ['BoltTool', 'LernaTool', 'PnpmTool', 'RootTool', 'RushTool', 'YarnTool', 'InvalidMonorepoError']
```

So `InvalidMonorepoError` is the only one at odds here.

> I suppose this is true although I do instanceof on Errors exported by e.g. AWS SDK without issue.

This is definitely an edge case that doesn't bite us too often but when it does... it hurts 😅 

> If we wanted to truly fix this then I'd recommend adding a static isInvalidMonorepoError(error: unknown) boolean onto the Error, and adding a property code: "InvalidMonorepoError" to the class to perform the check.

Yeah, that would work - at the expense of leaving that export in `@manypkg/tools`. Perhaps it's not worth trying to optimize the `import * as tools from '@manypkg/tools'` use case. I just thought that it would be kinda cool if we'd have that "trait".

packages/get-packages/package.json
```diff
@@ -7,6 +7,7 @@
     "@babel/runtime": "^7.20.6",
     "@changesets/types": "^4.0.1",
     "@manypkg/find-root": "^1.1.0",
+    "@manypkg/tools": "^1.0.0",
     "fs-extra": "^8.1.0",
     "globby": "^11.0.0",
     "read-yaml-file": "^1.1.0"
```
Andarist, line 13: This packages only depends on those now:
```suggestion
    "@babel/runtime": "^7.20.6",
    "@manypkg/find-root": "^1.1.0",
    "@manypkg/tools": "^1.0.0"
```

packages/tools/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@manypkg/tools",
+  "version": "1.0.0",
+  "license": "MIT",
+  "main": "dist/manypkg-tools.cjs.js",
+  "dependencies": {
+    "@babel/runtime": "^7.20.6",
+    "fs-extra": "^8.1.0",
+    "globby": "^11.0.0",
+    "jju": "^1.4.0"
```
Andarist, line 10: `PnpmTool` needs this:
```suggestion
    "jju": "^1.4.0",
    "read-yaml-file": "^1.1.0"
```

Andarist (approved):

PR #156, opened 2022-12-20
Suggested fixes to tsconfig, prettier config

elliot-nelson:
### SUMMARY

 - Move _prettier config_ into Prettier's dedicated config file, for ease of discovery
 - Add the `prettier-plugin-packagejson` prettier plugin
 - Turn off `useUnknownInCatchVariables` typescript rule

### DETAILS

I am open to pushback on this, but I think in general, using `package.json` as an unstructured grab-bag is less discoverable than using tool-specific configuration files. So I was hoping to split out `.prettierrc.js` into its own file in the root.

It would also be nice to add `prettier-plugin-packagejson` -- it's very opinionated, as it wraps the opinionated `sort-package-json` utility, but I've found it really nice to just give in and let VSCode auto-sort my whole packagejson every time I touch it.

Last, I think we (and everyone else) should turn off `useUnknownInCatchVariables`. Notably, even the [TypeScript codebase itself turns this off](https://github.com/microsoft/TypeScript/issues/42596) -- the extra safety is not worth the hassle until TS starts offering some better assumptions (like at least being able to specify `err: Error`).

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -8,8 +8,7 @@ module.exports = {
   // Avoid line breaks in markdown
   proseWrap: "never",
 
-  // Disabling search dirs and pointing directly at plugins
-  // makes it easier for VSCode and WebStorm to find them.
+  // Disabling search dirs and pointing directly at plugins to keep configuration explicit
   pluginSearchDirs: false,
   plugins: [require("prettier-plugin-packagejson")],
 };
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,15 @@
+// Documentation for this file: https://prettier.io/docs/en/configuration.html
+module.exports = {
+  // Use .gitattributes to manage newlines
+  endOfLine: "auto",
+
+  // Avoid line breaks in markdown
+  proseWrap: "never",
+
+  // Prettier plugins
+  //
+  // Disabling search dirs and pointing directly at folder paths relative to this config file
+  // makes it easier for VSCode and WebStorm to find your plugins.
+  pluginSearchDirs: [],
+  plugins: ["./node_modules/prettier-plugin-packagejson"]
+};
--- a/package.json
+++ b/package.json
@@ -7,6 +7,18 @@
     "packages/*",
     "test-gatsby"
   ],
+  "scripts": {
+    "postinstall": "preconstruct dev && manypkg check",
+    "release": "yarn preconstruct build && yarn changeset publish",
+    "test": "jest",
+    "test-gatsby": "cd test-gatsby && gatsby develop"
+  },
+  "jest": {
+    "watchPlugins": [
+      "jest-watch-typeahead/filename",
+      "jest-watch-typeahead/testname"
+    ]
+  },
   "dependencies": {
     "@babel/core": "^7.20.5",
     "@babel/preset-env": "^7.20.2",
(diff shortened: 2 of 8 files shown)


tsconfig.json
```diff
@@ -46,9 +46,10 @@
     // "typeRoots": [],                       /* List of folders to include type definitions from. */
     // "types": [],                           /* Type declaration files to be included in compilation. */
     // "allowSyntheticDefaultImports": true,  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
-    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
+    "esModuleInterop": true,                  /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */
     // "preserveSymlinks": true,              /* Do not resolve the real path of symlinks. */
     // "allowUmdGlobalAccess": true,          /* Allow accessing UMD globals from modules. */
+    "useUnknownInCatchVariables": false
```
Andarist, line 52: I was in that camp... until I got bit by some rogue library throwing a trash value. I won't say that I'm always writing code that adheres to this strict setting but I feel like it's a better choice to have it on. It's also super easy to opt out of it on the call site with:
```ts
} catch (err: any) {
```

and at that point, it's at least in plain sight that we are a little bit on the safe path which is an added bonus for me. This doesn't allow us to do this:
```ts
} catch (err: Error) {
```

but that's also a good thing for me - the cast becomes explicit if we want to reach for it.

And especially with the recent changes in TS 4.9 related to the `in` narrowing the `unknown` type this isn't that bad.

```suggestion
```

Not a hill I'm willing to die on but this one doesn't annoy me enough to turn it off.

package.json
```diff
@@ -22,26 +34,12 @@
...
-    ]
-  },
   "preconstruct": {
     "packages": [
       "packages/!(gatsby)*"
     ]
-  },
-  "prettier": {
-    "proseWrap": "never"
-  },
```
Andarist, line 40: I'm in the "explicit configs camp" too 👍

.prettierrc.js
```diff
@@ -0,0 +1,15 @@
...
+  // Use .gitattributes to manage newlines
+  endOfLine: "auto",
+
+  // Avoid line breaks in markdown
+  proseWrap: "never",
+
+  // Prettier plugins
+  //
+  // Disabling search dirs and pointing directly at folder paths relative to this config file
+  // makes it easier for VSCode and WebStorm to find your plugins.
```
Andarist, line 12: interesting - do you happen to have a link to some issues related to this? I'm just curious why the automatic plugin search tends to fail in those contexts

elliot-nelson: Now that you've challenged me on it, I suspect that this is a defense mechanism for Rush in particular.

If you have a Lerna repo or something and you use a root-level `package.json` to keep your prettier devDependency, which lives in a root-level `node_modules`, then I think 99.9% of the time VSCode would "do the right thing".

But in a Rush repo, there is no root package or node modules, and plugins are kept in a special folder; VSCode typically can't find this folder unless you specify exactly where it is. I can test out simplifying this for this repo since it probably would work just fine.

Andarist: Ah, ye - this might makes sense in the context of Rush if the location of plugins etc is not standard. That being said, I **prefer** an explicit list of plugins rather than one based on auto-search.

tsconfig.json
elliot-nelson: Fair enough! I'll remove this from PR

.prettierrc.js
```diff
@@ -0,0 +1,15 @@
...
+/** @type {import('prettier').Options} */
+module.exports = {
+  // Use .gitattributes to manage newlines
+  endOfLine: "auto",
+
+  // Avoid line breaks in markdown
+  proseWrap: "never",
+
+  // Disabling search dirs and pointing directly at plugins
+  // makes it easier for VSCode and WebStorm to find them.
```
Andarist, line 12: The latter is probably not true in our case so let's shorten this
```suggestion
  // Disabling search dirs and pointing directly at plugins to keep configuration explicit
```

elliot-nelson:
^ One note -- I don't know if this is normal or not -- but it looks like this PR doesn't run CI checks at all (I'd feel a bit safer if it was actually running CI on a change like this).

Andarist:
> ^ One note -- I don't know if this is normal or not -- but it looks like this PR doesn't run CI checks at all (I'd feel a bit safer if it was actually running CI on a change like this).

It's the whole mining prevention measure implemented by GitHub. For first-time contributors, the CI needs a maintainer's approval to run. I've approved this workflow - it should pass in a moment, I will land this and you won't have that problem again.

PR #165, opened 2023-01-23
Removed support for Bolt monorepos

Andarist:


elliot-nelson (approved):

elliot-nelson:
☠️ Bolt

lukebatchelor:
Ah, that's disappointing.

Any particular reason why @Andarist ?

Andarist:
I dont mind sypporting it - i just truly thought that it’s not worth it since its downloads are very low and the project seems to be completely stagnated. I’d accept a PR reverting this one

PR #167, opened 2023-01-23
Caller can override the list of supported tools

elliot-nelson:
### SUMMARY

 - Allow users of `find-root` and `get-packages` to provider their own list of tools.

### DETAILS

 - By default, current behavior is unchanged.
 - Users can provide an array of existing tool implementations as a second argument, passing in their own custom tools or reordering the default set of tools.
 - `RootTool` retains its special handling in the `find-root` package, but the user can choose to disable finding a single-package repo by omitting this tool from their list (this results in a different exception if no other matching monorepo is discovered).


packages/find-root/src/index.ts
```diff
@@ -20,34 +20,56 @@ import {
...
+ * is found, and return a `MonorepoRoot` object with the discovered directory and a
+ * corresponding monorepo `Tool` object.
+ *
+ * By default, all predefined `Tool` implementations are included in the search -- the
+ * caller can provide a list of desired tools to restrict the types of monorepos discovered,
+ * or to provide a custom tool implementation.
+ */
+export async function findRoot(
+  cwd: string,
+  tools?: Tool[]
```
Andarist, line 64: I'd probably go with an object argument right from the start:
```suggestion
  { tools = DEFAULT_TOOLS }: { tools?: Tool[] }
```

packages/find-root/src/index.ts
```diff
@@ -71,6 +93,10 @@ export async function findRoot(cwd: string): Promise<MonorepoRoot> {
     return monorepoRoot;
   }
 
+  if (!toolsToInclude.includes(RootTool)) {
+    throw new NoMatchingMonorepoFound(cwd);
+  }
```
Andarist, line 98: I'm not entirely sure without actually testing this further - but I think that this actually won't work. Imagine this

```js
/*
  packages/package-a/package.json
  package.json
  pnpm-lock.yaml
  pnpm-workspace.yaml
*/
findRoot(pkgA, {
  tools: [PnpmTool, RootTool]
})
```

If I'm not mistaken - this would determine `packages/package-a` based on the `RootTool` being included in the list but the intention of the caller was clearly to find the pnpm root here.

Calling this properly, while allowing `RootTool`, would require something like this:
```ts
const root = (await findRoot(pkgA, { tools: [PnpmTool] })) || (await findRoot(pkgA, { tools: [RootTool] }))
```

Andarist: Ah, so it actually doesn't work like that since `RootTool.isMonorepoRoot` always returns `false` and we just have a special path for it in the `findRoot`. 

Disregard that comment, I also missed that this PR adds `RootTool` to the `DEFAULT_TOOLS`

packages/find-root/src/index.ts
```diff
@@ -20,34 +20,56 @@ import {
...
 
-export async function findRoot(cwd: string): Promise<MonorepoRoot> {
+/**
+ * Given a starting folder, search that folder and its parents until a supported monorepo
+ * is found, and return a `MonorepoRoot` object with the discovered directory and a
+ * corresponding monorepo `Tool` object.
+ *
+ * By default, all predefined `Tool` implementations are included in the search -- the
+ * caller can provide a list of desired tools to restrict the types of monorepos discovered,
+ * or to provide a custom tool implementation.
```
Andarist, line 60: This sounds handy - especially the ability to provide a custom tool. I wonder though if the ability to restrict the types of discoverable monorepos isn't a YAGNI thing. OTOH, if we allow customizing a single tool, then it's really not a big problem for us to allow a list of custom tools. And at that point, "restricting" isn't really different from "custom list"

packages/find-root/src/index.ts
```diff
@@ -20,34 +20,56 @@ import {
  * monorepo implementations first, with tools based on custom file schemas
  * checked last.
  */
-const defaultOrder: Tool[] = [
+export const DEFAULT_TOOLS = [
```
Andarist, line 23: I'd probably avoid exporting this for as long as it's not requested by the users. It's hard to predict how consumers use this and it could lock us from adding new items to this list.

Andarist: Do you have any particular code in mind that would utilize this?

elliot-nelson: @Andarist By exporting the tool list I think it allows the caller to decide between two approaches:

```javascript
// explicitly limiting to my desired tools
findRootSync('.', [MyCustomTool, PnpmTool, YarnTool]);

// i just don't want to allow RootTool
findRootSync('.', DEFAULT_TOOLS.filter(tool => tool !== RootTool));
```

You can simulate the second option by just copying + pasting the list into your code and removing the ones you dislike; but, any future releases of tools won't automatically be used, you're on the hook to keep up-to-date.

packages/find-root/src/index.ts
elliot-nelson: Hah, YAGNI is definitely a concern, but I think my arguments for this approach would be:

 - I _definitely_ want to be able to provide my own custom tool from the caller.
 - I _definitely_ want to be able to specify a desired pre-baked tool.
 - But by default, I want to use a list of supported tools.
 - So... a customizable array of tools is a supserset of these requirements, but it also seems like the simplest approach that covers them all.

packages/find-root/src/index.ts
elliot-nelson: That's fair, I had written it this way just because the signature was identical to `get-packages`, which I thought made it maybe easy to see that they are the same signature.

(In the case of `get-packages`, I do want to pass along the undefined, so the default order isn't locked in from two different code paths.)

Andarist: Yeah, totally - `get-packages` should use the default param for this. Otherwise those signatures could still be "the same"

packages/find-root/src/index.ts
Andarist: 👍

packages/find-root/src/index.ts
elliot-nelson: OK -- I'm willing to keep it internal for now, we can revisit if I end up with a definitive use case :)

elliot-nelson: (I've made that change and also the other feedback.)

packages/find-root/src/index.ts
```diff
@@ -19,28 +19,54 @@ import {
...
+ * is found, and return a `MonorepoRoot` object with the discovered directory and a
+ * corresponding monorepo `Tool` object.
+ *
+ * By default, all predefined `Tool` implementations are included in the search -- the
+ * caller can provide a list of desired tools to restrict the types of monorepos discovered,
+ * or to provide a custom tool implementation.
+ */
+export async function findRoot(
+  cwd: string,
+  tools: Tool[] = DEFAULT_TOOLS
```
Andarist, line 62: What do you think about making this argument an object? I'm worried that using a position argument here might become cumbersome in the future if we ever end up wanting to add a new parameter

elliot-nelson: Kick off an options interface? Makes sense to me, I'll put that in this weekend.

elliot-nelson:
Latest push: convert from a second `tools` parameter to new options interface, for future extension.

packages/find-root/src/index.ts
```diff
@@ -19,28 +19,65 @@ import {
...
+ */
+export interface FindRootOptions {
+  /**
+   * Override the list of monorepo tool implementations that are used during the search.
+   */
+  tools?: Tool[];
+}
+
+/**
+ * Given a starting folder, search that folder and its parents until a supported monorepo
```
Andarist, line 62: We probably should stick to "directory" over "folder" since in the APIs we use `dir`, `relativeDir` etc

.changeset/many-pumpkin-behavior.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/find-root": patch
```
Andarist, line 2:
```suggestion
"@manypkg/find-root": minor
```

.changeset/odd-lies-perform.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/get-packages": patch
```
Andarist, line 2:
```suggestion
"@manypkg/get-packages": minor
```

Issue #172, opened 2023-02-23
Please restore Bolt support


steve-taylor:
I realise Bolt didn't win the monorepo package manager war, but there are still people using it (e.g. me). For me, its killer features are enforcement of uniform dependency versions, speed and simplicity. I've tried them all and settled on a toolkit that included Bolt, Changesets and Manypkg.

Unfortunately, Changesets uses Manypkg to get workspaces, so Bolt support in Changesets is going away soon.

Can you please restore Bolt support? Alternatively, will you accept a well-crafted pull request that restores Bolt support?

Andarist:
> Can you please restore Bolt support? Alternatively, will you accept a well-crafted pull request that restores Bolt support?

ye, i think you should be able to revert my PR that removed it

steve-taylor:
Thanks! PR coming soon. While I'm at it, bolt seems to support exclusion patterns, whereas manypkg doesn't. So I'll fix that.

PR #174, opened 2023-04-10
Restore support for Bolt monorepos

steve-taylor:
This addresses #172 by reverting #165, with some minor modifications that I've annotated in comments.

### DIFF
--- a/packages/find-root/src/index.ts
+++ b/packages/find-root/src/index.ts
@@ -6,6 +6,7 @@ import {
   Tool,
   RootTool,
   MonorepoRoot,
+  BoltTool,
   LernaTool,
   PnpmTool,
   RushTool,
@@ -21,6 +22,7 @@ import {
  */
 const DEFAULT_TOOLS: Tool[] = [
   YarnTool,
+  BoltTool,
   PnpmTool,
   LernaTool,
   RushTool,
@@ -32,6 +34,7 @@ const isNoEntryError = (err: unknown): boolean =>
 
 export class NoPkgJsonFound extends Error {
   directory: string;
+
   constructor(directory: string) {
     super(`No package.json could be found upwards from directory ${directory}`);
     this.directory = directory;
@@ -40,6 +43,7 @@ export class NoPkgJsonFound extends Error {
 
 export class NoMatchingMonorepoFound extends Error {
   directory: string;
+
   constructor(directory: string) {
     super(
       `No monorepo matching the list of supported monorepos could be found upwards from directory ${directory}`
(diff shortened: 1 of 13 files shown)


packages/get-packages/README.md
```diff
@@ -1,21 +1,35 @@
...
+    isMonorepoRootSync(directory: string): boolean;
+    getPackages(directory: string): Promise<Packages>;
+    getPackagesSync(directory: string): Packages;
+}
+
+interface Package {
+    packageJson: PackageJSON;
+    dir: string;
+    relativeDir: string;
+}
```
steve-taylor, line 27: I added `Tool` and `Package` here as `Packages` (below) refers to them. Not sure if that's needed/wanted. Happy to remove.

packages/find-root/src/index.ts
```diff
@@ -32,6 +34,7 @@ const isNoEntryError = (err: unknown): boolean =>
 
 export class NoPkgJsonFound extends Error {
   directory: string;
+
```
steve-taylor, line 37: Prettier

packages/find-root/src/index.ts
```diff
@@ -40,6 +43,7 @@ export class NoPkgJsonFound extends Error {
 
 export class NoMatchingMonorepoFound extends Error {
   directory: string;
+
```
steve-taylor, line 46: Prettier

packages/get-packages/src/index.ts
```diff
@@ -6,6 +10,7 @@ export type { Tool, Package, Packages } from "@manypkg/tools";
 
 export class PackageJsonMissingNameError extends Error {
   directories: string[];
+
```
steve-taylor, line 13: Prettier

packages/get-packages/README.md
Andarist: perhaps we should refer to them as `import('@manypkg/tools').Tool` etc? This way we wouldn't have to worry about syncing those definitions in READMEs and it would be clear from where they are coming from here

packages/get-packages/src/index.ts
```diff
@@ -1,3 +1,7 @@
+// This is a modified version of the package-getting in bolt
+// It supports yarn workspaces as well, and can fall back through
+// several options
```
Andarist, line 3: I feel that this has been refactored so much lately that this comment is no longer applicable
```suggestion
```

packages/find-root/src/index.ts
```diff
@@ -21,6 +22,7 @@ import {
  */
 const DEFAULT_TOOLS: Tool[] = [
   YarnTool,
+  BoltTool,
```
Andarist, line 25: would it work if we'd put `BoltTool` at the end?

steve-taylor: Yeah, it still works. I moved it down. It's still above `RootTool` as it seems like`RootTool` should always be last.

Anything else or are we good to merge?

PR #175, opened 2023-04-22
enable possibility to end name with "/" for scoped packages

dohomi:
This would be very helpful to still upgrade scoped packages like `@scoped-name/` which currently fails:
```
manypkg upgrade @scoped-name/
```


dohomi:
Example:
```
manypkg upgrade @tamagui/
```
currently does not work. 
```
manypkg upgrade @tamagui
```
works but it also upgrades `@tamagui-extras` which is not intended.

emmatown:
Could you update so this writing `@scope` in the CLI looks for packages starting with `@scope/` rather than `@scope`? That was the original intent.

Andarist:
Closed by https://github.com/Thinkmill/manypkg/pull/206

PR #183, opened 2023-06-27
Update and move `@types/node` to dev updates

### LINKED ISSUE #118: move `@types/node` to `devDependencies`?
As title.

It is not a good practice to install `@types/node` as dependency AFAIK.

askoufis:
`@types/node` should not be a dependency for `find-root` since it doesn't expose any of the node types in its API, so I've moved it to a dev dep. I also updated the version to the last Node 16 types version. I chose Node 16 as it's [the same version used in CI](https://github.com/Thinkmill/manypkg/blob/main/.github/workflows/main.yml#L19).

Fixes #118.

### DIFF
--- a/.changeset/node-types-lag.md
+++ b/.changeset/node-types-lag.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/find-root": patch
+---
+
+Remove `@types/node` dependency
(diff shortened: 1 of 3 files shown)


.changeset/node-types-lag.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/find-root": patch
+---
+
+Remove `@types/node` dependency
```
Andarist, line 5:
```suggestion
Moved `@types/node` out of `dependencies` to `devDependencies`
```

PR #185, opened 2023-07-03
Added repository references to packages

vergilfromadyen:


packages/gatsby-source-workspace/package.json
```diff
@@ -6,6 +6,11 @@
     "gatsby-node.js",
     "index.js"
   ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Thinkmill/manypkg.git",
+    "directory": "packages/gatsby-source-workspace"
+  },
```
Andarist, line 13: usually I was using just smth like:
```suggestion
  "repository": "https://github.com/Thinkmill/manypkg/tree/main/packages/gatsby-source-workspace",
```

is there any practical difference between those two? Are both supported in the same way by various tools that might be interested in this metadata?

vergilfromadyen: I suppose the difference would be that in this case the URL contains the name of the branch. Right now `main` seems safe but the world already renamed `master` to `main` so this wouldn't have been future-proof if introduced in the past. Right now it seems unlikely that we'll do a similar change anytime soon but still it doesn't seem necessary to include branch names in the repo reference.

But really the main purpose is to have the NPM registry reference this repo so automated tools can find the repo from a package name.

Andarist: Do you know any package using this pattern that is published to npm? I vaguely remember investigating this in the past (over a year ago or something) and IIRC I was very surprised that npm site didn't actually support this despite it being documented in the npm docs.

vergilfromadyen: Sure - here's [jest](https://github.com/jestjs/jest/blob/main/packages/jest-diff/package.json) and [here's an example straight from the docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository:~:text=If%20the%20package.json%20for%20your%20package%20is%20not%20in%20the%20root%20directory%20(for%20example%20if%20it%20is%20part%20of%20a%20monorepo)%2C%20you%20can%20specify%20the%20directory%20in%20which%20it%20lives%3A)

benmccann:
@vergilfromadyen I didn't notice this PR and sent the same one here: https://github.com/Thinkmill/manypkg/pull/208

Mine was merged, so you can close this now. Thanks for sending this!

benmccann:
@emmatown while you're going through things, this PR could be closed

Issue #72, opened 2020-07-30
Prefer workspace protocol when using pnpm 


ifiokjr:
## Description

When `pnpm` encounters a package in a workspace which has a lower version than the most recently published on the `npm` registry it defaults to using the registry version. 

I opened an [issue](https://github.com/pnpm/pnpm/issues/2725#issuecomment-665512775) and the proposed solution was to use [workspace protocols](https://github.com/pnpm/pnpm/issues/2725#issuecomment-665512775).

Unfortunately, `manypkg fix` removes all workspace protocols when run.

My proposal is not just to ignore workspace protocols when using `pnpm` but actually to prefer them, since it's the recommended way to prevent a hard situation to debug. It took me a while to realize why none of my changes were having an impact.

maraisr:
@mitchellhamilton @Andarist would be so awesome to get an answer here. More than happy to explore at a solution as well.

Andarist:
I assume that the proposal here is for `fix` to enforce the workspace protocol (when used with `pnpm`). What about `check`? Should it allow both? Warn when it encounters non-workspace dependency? I've seen people mixing workspace package with non-workspace ones intentionally (the cost of simultaneous upgrades in large monorepos was too high for them). OTOH the "Internal mismatch" rule explicitly forbids this - however, with workspace protocol one can have a more explicit and greater control over this. So maybe in this context it's not worth enforcing this rule so strictly - without the workspace protocol (yarn v1, npm?) this was much a bigger issue because one couldn't control this in such a way.

maraisr:
@Andarist what is the go here? Can I raise a pr to fix something? Just need to know what needs to happen.

chaance:
I'm also running into this issue. Has anyone found a workaround since the issue was originally raised?

mattpocock:
Wanted to flag this since it makes manypkg kind of unusable with pnpm. Would you accept a PR on this?

Andarist:
Yeah, I think a PR for this would be accepted. At the very least we should allow those ranges and don't raise the `INTERNAL_MISMATCH` error. We should have tests for `workspace:*`, `workspace:^`, `workspace:~` and `workspace:1.2.3`. Because `workspace:^` and `workspace:~` ranges are valid a simple `.slice` like mentioned by @chaance might not be enough - those could be mapped to just `*` though.

[...]

PR #186, opened 2023-07-22
Start support for `workspace:` protocol for internal packages

webpro:
This PR starts support for the `workspace:` protocol and does many of the things as discussed in #72 to make this tool usable with pnpm.

I think the code is pretty self-explanatory, also added a bunch of test for coverage.

Explicitly opts out of trying to fix errors, because I'm not sure whether that's 1) safe to do and/or 2) desired in all cases. I know in my own test workloads it's not what I want.

An option could be added to prefer the `workspace:` protocol but that's also outside the scope of this PR.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,5 +6,6 @@
   ],
   "commit": false,
   "linked": [],
-  "access": "public"
+  "access": "public",
+  "baseBranch": "main"
 }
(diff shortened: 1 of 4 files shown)


Stanzilla:
@emmatown @Andarist can someone take a look, please? pnpm support would be very nice to have

antoniopresto:
Does the maintainer hate pnpm?  😵‍💫

VanTanev:
We would also love to have this supported out of the box in manypkg

emmatown:
Thanks so much for this PR and sorry for taking so long on this, I wanted to take a slightly different approach to what's in this PR so I've added support for the `workspace:` protocol in #204

PR #188, opened 2023-09-13
feat: detect package manager for fix cmd

ryoid:
Improve the provided fix command for convenience.

```sh
> manypkg check

☔️ error X dependencies are unsorted, this can cause large diffs when packages are added, resulting in dependencies being sorted
☔️ info the above errors may be fixable with yarn manypkg fix
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -16,6 +16,7 @@
     "@manypkg/get-packages": "^2.2.0",
     "chalk": "^2.4.2",
     "detect-indent": "^6.0.0",
+    "detect-package-manager": "^3.0.1",
     "find-up": "^4.1.0",
     "fs-extra": "^8.1.0",
     "normalize-path": "^3.0.0",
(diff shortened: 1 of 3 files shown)


KalleV (approved):

benmccann:
I would recommend using package-manager-detector instead

https://npmgraph.js.org/?q=detect-package-manager - 16 dependencies 
https://npmgraph.js.org/?q=package-manager-detector - 0 dependencies 

https://www.npmjs.com/package/detect-package-manager - declining usage. no updates in past year
https://www.npmjs.com/package/package-manager-detector - more weekly downloads

bluwy:
I'm not sure if it's worth adding a dependency for a single log though. I think shortening as `manypkg fix` is enough as if you're already executing `manypkg check`, then you'd know how to execute `manypkg fix` too. Otherwise a more general solution is to use `npx manypkg fix`.

PR #192, opened 2023-11-30
fix: miss "types" property

zouguangxian:


Andarist:
This isn't strictly needed. TypeScript is able to locate our types even without this.

PR #194, opened 2024-01-12
fix: allow tags in `npm:x`-deps

### LINKED ISSUE #193: Bug: `npm:x@y` doesn't work for tags
Heya, super grateful for this lib, it keeps things tidy 🧹

In my case now, it seems to make things _too tidy_, where it doesn't allow me to use `@tag`

I added a dep in my monorepo for e.g. `"@trpc/server": "npm:@trpc/server@next",` (note it has a tag and not a version at the end). Manypkg does not like that.

```
☔️ error @examples/standalone-server has a dependency on @trpc/server@npm:@trpc/server@next but the version of @trpc/server in the repo is 10.45.0 which is not within range of the depended on version, please update the dependency version
```

See https://github.com/trpc/trpc/pull/5315:

CleanShot 2024-01-11 at 21 34 01@2x

KATT:
Closes #193

### DIFF
--- a/packages/cli/src/checks/INTERNAL_MISMATCH.ts
+++ b/packages/cli/src/checks/INTERNAL_MISMATCH.ts
@@ -25,8 +25,8 @@ export default makeCheck<ErrorType>({
           
           if (
             dependencyWorkspace !== undefined &&
-            !semver.satisfies(dependencyWorkspace.packageJson.version, range) && 
-            !range.startsWith("npm:")
+            !range.startsWith("npm:") &&
+            !semver.satisfies(dependencyWorkspace.packageJson.version, range)
           ) {
             errors.push({
               type: "INTERNAL_MISMATCH",
--- a/packages/cli/src/checks/INTERNAL_MISMATCH.ts
+++ b/packages/cli/src/checks/INTERNAL_MISMATCH.ts
@@ -22,6 +22,7 @@ export default makeCheck<ErrorType>({
         for (let depName in deps) {
           let range = deps[depName];
           let dependencyWorkspace = allWorkspaces.get(depName);
+          
           if (
             dependencyWorkspace !== undefined &&
             !semver.satisfies(dependencyWorkspace.packageJson.version, range)
--- a/packages/cli/src/checks/INTERNAL_MISMATCH.ts
+++ b/packages/cli/src/checks/INTERNAL_MISMATCH.ts
@@ -25,7 +25,8 @@ export default makeCheck<ErrorType>({
           
           if (
             dependencyWorkspace !== undefined &&
-            !semver.satisfies(dependencyWorkspace.packageJson.version, range)
+            !semver.satisfies(dependencyWorkspace.packageJson.version, range) && 
+            !range.startsWith("npm:")
           ) {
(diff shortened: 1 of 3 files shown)


packages/cli/src/checks/INTERNAL_MISMATCH.ts
```diff
@@ -22,9 +22,11 @@ export default makeCheck<ErrorType>({
         for (let depName in deps) {
           let range = deps[depName];
           let dependencyWorkspace = allWorkspaces.get(depName);
+          
           if (
             dependencyWorkspace !== undefined &&
-            !semver.satisfies(dependencyWorkspace.packageJson.version, range)
+            !semver.satisfies(dependencyWorkspace.packageJson.version, range) && 
+            !range.startsWith("npm:")
```
Andarist, line 29:
```suggestion
            !range.startsWith("npm:") &&
            !semver.satisfies(dependencyWorkspace.packageJson.version, range)
```

packages/cli/src/checks/__tests__/INTERNAL_MISMATCH.ts
```diff
@@ -85,4 +85,20 @@ describe("internal mismatch", () => {
       ]
     `);
   });
+
+  // regression test for https://github.com/Thinkmill/manypkg/issues/193
+  it.each(["npm:pkg-1@sometag", "npm:someorg@pkg-1@sometag"])(
```
Andarist, line 90: Shouldn't this be something like this?
```suggestion
  it.each(["npm:pkg-1@sometag", "npm:@someorg/pkg-1@sometag"])(
```

packages/cli/src/checks/INTERNAL_MISMATCH.ts
Andarist: I know about the existence of `npm:` ranges and used them a couple of times... but damn, I can't find any docs that would mention them 😅 

In general, we likely should relax this check more and always allow things like `file:`, `link:`, etc. We don't necessarily have to do it here but maybe a simple regex that would allow any protocol would do the job here? cc @emmatown do you have any opinions about this one?

KATT: It's mentioned here: https://github.com/Thinkmill/manypkg?tab=readme-ov-file#ignoring-this-rule

KATT: I did the change you requested here

Andarist: I meant that I can't find this in the npm/yarn/pnpm docs. pnpm has a mention of this in one of examples but it doesn't even explain what it does 😅

emmatown (approved):

packages/cli/src/checks/INTERNAL_MISMATCH.ts
emmatown: > In general, we likely should relax this check more and always allow things like file:, link:, etc. We don't necessarily have to do it here but maybe a simple regex that would allow any protocol would do the job here?

This sounds good to me

PR #201, opened 2024-01-19
Fix start directory for `getPackages`

### LINKED ISSUE #200: `getPackages` fails when given a monorepo subdirectory instead of root directory.
`getPackages` expects `dir` argument to be the monorepo root but its description says otherwise. 

https://github.com/Thinkmill/manypkg/blob/52d5dcb469b1d6781f56cff15b3556c8856ce8ca/packages/get-packages/src/index.ts#L26-L40

Structure:

```
├──📄 package.json
├──📁 packages
│	├──📁 pkg-a
│	│	├──📄 index.mjs
│	│	├──📄 package.json
│	├──📁 cli
│	│	├──📄 bin.mjs
│	│	├──📄 package.json
├──📁 apps
│	├──📁 app-a
│	│	├──📄 package.json
│	│	├──📁 src
```

- `cli` package is consumed by monorepo root

### Steps:
- Open Codesandbox Devbox - https://codesandbox.io/p/devbox/manypkg-get-packages-bug-5zdymd
- Sign in or download the devbox
- Install the dependencies - `yarn` or you can also set it up using `pnpm`
- In the root of the monorepo, run `yarn dev-cli`
- Verify the outfit from the CLI  package
- `cd packages/pkg-a` and run `yarn dev-cli` again

### Expected Behavior:
The command should be successful and should be able to output the monorepo packages

### Actual Behavior:
It throws an error - `TypeError: Cannot read properties of undefined (reading 'packages')`

Should it be

```ts
const monorepoRoot: MonorepoRoot = await findRoot(dir, options);
const packages: Packages = await monorepoRoot.tool.getPackages(monorepoRoot.rootDir);
//                                                             ^^^^^^^^^^^^^^^^^^^^
```
instead of

```ts
const monorepoRoot: MonorepoRoot = await findRoot(dir, options);
const packages: Packages = await monorepoRoot.tool.getPackages(dir);
//                                                             ^^^
```

manzoorwanijk:
Fixes #200

This PR updates `getPackages` to start looking from the monorepo root, instead of the passed `dir`

### DIFF
--- a/.changeset/breezy-readers-chew.md
+++ b/.changeset/breezy-readers-chew.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/get-packages": patch
+---
+
+Fixed the error for getPackages when given a non-root directory
--- a/packages/get-packages/src/index.test.ts
+++ b/packages/get-packages/src/index.test.ts
@@ -1,4 +1,5 @@
 import fixturez from "fixturez";
+import path from 'node:path';
 import { getPackages, getPackagesSync } from "./";
 
 const f = fixturez(__dirname);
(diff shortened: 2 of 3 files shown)


emmatown (approved):
Thanks!

PR #204, opened 2024-03-15
Allow dependecies to use the `workspace:` protocol and add `"workspaceProtocol": "require"` config option

### LINKED ISSUE #72: Prefer workspace protocol when using pnpm

### LINKED ISSUE #147: support/configure allowing `workspace:*`
Would it be possible to allow for `workspace:*` to be used as version?

I use this for internal packages that I don't publish in [create-t3-turbo](https://github.com/t3-oss/create-t3-turbo). It works on main that is using npm but in [this PR where I'm migrating to PNPM](https://github.com/t3-oss/create-t3-turbo/pull/27) manypkg fails with [this error](https://app.warp.dev/block/4gItGcNTOGThzfmzTIv7tp).

emmatown:
Note that this doesn't enforce that what is after the `workspace:` protocol is correct since package managers will do that themselves.

Fixes #72
Fixes #147


jrolfs:
Thank you!

butterybread:
Awesome! 🎉

Issue #148, opened 2022-10-06
support asterix in upgrade


dohomi:
Hello
I have an edge case where I would like to run 
```
manypkg upgrade mypackage *
```
to run and upgrade all `mypackage` to * because its part of the monorepo. Its an edge case because the `mypackage` name collides with the upgrade of another package with the same starting name.

Is that possible?

dohomi:
I prepared a PR to support scoped package names ending with a "/" https://github.com/Thinkmill/manypkg/pull/175

Andarist:
Closed by https://github.com/Thinkmill/manypkg/pull/206

PR #206, opened 2024-04-13
Fix Upgrade to Properly Update Scoped Packages

bdkopen:
### Issue
Running `manypkg upgrade @scope` currently upgrades packages formatted as `@scope/package` and `@scope-extra/package`. This is problematic because it should not be updating `@scope-extra/` packages.

### Solution
Properly upgrade package so `manypkg upgrade @scope` only upgrades `@scope/` packages and `manypkg upgrade @scope-extras` only upgrades `@scope-extras/` packages.

### Related
- Addresses #148
- Fixes the PR changes requested in #175

### DIFF
--- a/.changeset/weak-icons-invent.md
+++ b/.changeset/weak-icons-invent.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/cli": patch
+---
+
+Fix `manypkg upgrade @scope` upgrading packages like `@scope-something/a`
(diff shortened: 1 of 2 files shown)


Andarist (approved):
LGTM (cc @emmatown ). If this doesn't get an extra review within a week - please feel free to ping me and I'll land it then.

bdkopen:
Thanks @Andarist and @emmatown!

Should #148 and #175 be closed now that this is released?

PR #210, opened 2024-07-07
remove fs-extra from find-root

benmccann:


Andarist:
Let's merge this with https://github.com/Thinkmill/manypkg/pull/212 . Both PRs are about the same thing - I don't want to treat them as separate issues even if they touch separate packages

PR #212, opened 2024-07-07
remove fs-extra from dependencies

benmccann:
removes 4 transitive dependencies from the project: https://npmgraph.js.org/?q=fs-extra@8.1.0

### DIFF
--- a/packages/find-root/src/index.ts
+++ b/packages/find-root/src/index.ts
@@ -1,6 +1,7 @@
 import findUp, { sync as findUpSync } from "find-up";
 import path from "path";
 import fs from "fs";
+import fsp from "fs/promises";
 
 import {
   Tool,
@@ -115,7 +116,7 @@ export async function findRoot(
   let rootDir = await findUp(
     async (directory) => {
       try {
-        await fs.promises.access(path.join(directory, "package.json"));
+        await fsp.access(path.join(directory, "package.json"));
         return directory;
       } catch (err) {
         if (!isNoEntryError(err)) {
--- a/packages/find-root/src/index.ts
+++ b/packages/find-root/src/index.ts
@@ -1,6 +1,6 @@
 import findUp, { sync as findUpSync } from "find-up";
 import path from "path";
-import fs from "fs-extra";
+import fs from "fs";
 
 import {
   Tool,
@@ -115,7 +115,7 @@ export async function findRoot(
   let rootDir = await findUp(
     async (directory) => {
       try {
-        await fs.access(path.join(directory, "package.json"));
+        await fs.promises.access(path.join(directory, "package.json"));
         return directory;
       } catch (err) {
         if (!isNoEntryError(err)) {
--- a/packages/tools/src/BoltTool.ts
+++ b/packages/tools/src/BoltTool.ts
@@ -1,5 +1,5 @@
 import path from "path";
-import fs from "fs-extra";
+import fs from "fs";
 
 import { Tool, PackageJSON, Packages, InvalidMonorepoError } from "./Tool";
 import {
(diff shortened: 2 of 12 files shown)


packages/tools/src/BoltTool.ts
```diff
@@ -18,9 +18,9 @@ export const BoltTool: Tool = {
 
   async isMonorepoRoot(directory: string): Promise<boolean> {
     try {
-      const pkgJson = (await fs.readJson(
+      const pkgJson = JSON.parse((await fs.promises.readFile(
```
Andarist, line 21: is `fs.promises` supported by node 14? this is our declared supported version in package.json. Does this version support `"fs/promises"`? I prefer the latter over `fs.promises`

benmccann: added in node 10: https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_readfile_path_options

packages/find-root/src/index.ts
```diff
@@ -115,7 +115,7 @@ export async function findRoot(
   let rootDir = await findUp(
     async (directory) => {
       try {
-        await fs.access(path.join(directory, "package.json"));
+        await fs.promises.access(path.join(directory, "package.json"));
```
Andarist, line 118: let's prefer using `fs/promises` throughout this PR

packages/tools/src/LernaTool.ts
```diff
@@ -59,10 +55,10 @@ export const LernaTool: Tool = {
     const rootDir = path.resolve(directory);
 
     try {
-      const lernaJson = await fs.readJson(path.join(rootDir, "lerna.json"));
-      const pkgJson = (await fs.readJson(
+      const lernaJson = await readJson(rootDir, "lerna.json") as LernaJson;
+      const pkgJson = JSON.parse((await fs.promises.readFile(
```
Andarist, line 59: this could use the util that you introduced, right?

packages/tools/src/YarnTool.ts
Andarist, line 1: this file could use the utils that you introduced too

benmccann: done

packages/tools/src/LernaTool.ts
benmccann: yep. done

packages/find-root/src/index.ts
benmccann: done

packages/tools/src/expandPackageGlobs.ts
```diff
@@ -26,19 +26,21 @@ export async function expandPackageGlobs(
   const discoveredPackages: Array<Package | undefined> = await Promise.all(
     directories.map((dir) =>
       fs
-        .readJson(path.join(dir, "package.json"))
+        .promises
```
Andarist, line 29: rogue `fs.promises` here :P

packages/tools/src/expandPackageGlobs.ts
```diff
@@ -68,9 +70,9 @@ export function expandPackageGlobsSync(
   const discoveredPackages: Array<Package | undefined> = directories.map(
     (dir) => {
       try {
-        const packageJson: PackageJSON = fs.readJsonSync(
+        const packageJson: PackageJSON = JSON.parse(fs.readFileSync(
```
Andarist, line 73: this could use the util that u introduced

packages/tools/src/expandPackageGlobs.ts
Andarist: In general, this thing here could be rewritten to make use of the util that u introduced. But I get that it's not really the focus of this PR so I'm fine leaving it out from here.

benmccann: Ah, thanks. I missed this one with my search since it was split into two lines. I've just left it written in the existing style for now to avoid changing too much and avoid introducing any accidental changes

benmccann:
@Andarist thanks for all the reviews on this. I've addressed your comments. Let me know if you had any other feedback

benmccann:
@Andarist just a gentle ping. I'd love to get this in given all the review you've done on it already. thanks!

emmatown (approved):

PR #213, opened 2024-07-07
remove six yaml-related dependencies

benmccann:
Replaces https://npmgraph.js.org/?q=read-yaml-file@1.1.0 with https://npmgraph.js.org/?q=js-yaml

`read-yaml-file` barely does anything and is just tiny wrapper around `js-yaml` that calls `strip-bom`. However, `strip-bom` is unnecessary as `js-yaml` now strips the bom character by itself: https://github.com/nodeca/js-yaml/blob/0d3ca7a27b03a6c974790a30a89e456007d62976/lib/loader.js#L1665

### DIFF
--- a/packages/tools/src/PnpmTool.ts
+++ b/packages/tools/src/PnpmTool.ts
@@ -13,18 +13,11 @@ import {
   expandPackageGlobsSync,
 } from "./expandPackageGlobs";
 
-function stripBom(s: string) {
-	// Catches EFBBBF (UTF-8 BOM) because the buffer-to-string
-	// conversion translates it to FEFF (UTF-16 BOM).
-	return s.charCodeAt(0) === 0xFEFF ? s.slice(1) : s;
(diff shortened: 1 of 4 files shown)


packages/tools/src/PnpmTool.ts
```diff
@@ -1,10 +1,9 @@
 import path from "path";
-import readYamlFile, { sync as readYamlFileSync } from "read-yaml-file";
+import yaml from "js-yaml";
```
Andarist, line 2: `read-yaml-file` - what was its supported node engines' range (for the specific version that we were using here)? and what is the range of the new package?

benmccann: `read-yaml-file` pulled in `js-yaml` as a dependency, so we were using `js-yaml` before as well. I checked and we're using v4 of `js-yaml` both before and after this PR so the supported version ranges shouldn't change at all

PR #214, opened 2024-07-08
switch to `fast-glob` to remove 6 dependencies

benmccann:
https://npmgraph.js.org/?q=globby - 23 dependencies
https://npmgraph.js.org/?q=fast-glob - 17 dependencies

### DIFF
--- a/.changeset/angry-houses-kiss.md
+++ b/.changeset/angry-houses-kiss.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/tools": patch
+---
+
+Replace `globby` with `fast-glob` to remove extra dependencies
--- a/packages/tools/package.json
+++ b/packages/tools/package.json
@@ -9,10 +9,10 @@
   "license": "MIT",
   "main": "dist/manypkg-tools.cjs.js",
   "dependencies": {
+    "fast-glob": "^3.3.2",
     "fs-extra": "^8.1.0",
     "jju": "^1.4.0",
-    "read-yaml-file": "^1.1.0",
-    "tinyglobby": "^0.2.2"
+    "read-yaml-file": "^1.1.0"
   },
   "devDependencies": {
     "@types/jju": "^1.4.2",
--- a/packages/tools/package.json
+++ b/packages/tools/package.json
@@ -9,10 +9,10 @@
   "license": "MIT",
   "main": "dist/manypkg-tools.cjs.js",
   "dependencies": {
-    "fast-glob": "^3.3.2",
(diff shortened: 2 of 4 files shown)


Andarist:
why a change from `fast-glob` to `tinyglobby`? What features do we lose in either case?

benmccann:
`fast-glob` has 17 dependencies (https://npmgraph.js.org/?q=fast-glob) whereas `tinyglobby` has only 2

I'm not sure I understand your question about features. `tinyglobby` supports all of the features we're using here (and all of the tests pass, which helps demonstrate that). If there's some particular feature that `fast-glob` has that you think we may use in the future I can check about getting that added to `tinyglobby`. We can always switch to `fast-glob` if needed in the future, but right now that seems pretty hypothetical. The use of globbing functionality here is pretty minimal

Andarist:
Not all glob libraries are compatible in terms of the patterns they support. So I wonder if any user-provided pattern can change meaning with this change. I'm also not to keen on changing this for a library that currently has 67 weekly downloads.

benmccann:
> Not all glob libraries are compatible in terms of the patterns they support. So I wonder if any user-provided pattern can change meaning with this change. 

That's a great call out and something I should have mentioned in the PR description. There is a very tiny change here, which is that numerical expansion within braces (e.g. `{01..03}`) is not supported with this new library. Besides that, all other glob functionality is the same since both `globby` and `tinyglobby` use picomatch under the covers for the majority of matching

>  I'm also not to keen on changing this for a library that currently has 67 weekly downloads.

It is a very new package. However, it has been adopted by a number of projects already. tinyglobby has now had over 30,000 downloads in its first week

emmatown (changes requested):
I agree with @Andarist here on using `fast-glob` over `tinyglobby` here for both the fact that `fastglobby` is still quite new and `fast-glob` is what at least [pnpm](https://github.com/pnpm/pnpm/blob/7fa4599dbc8a92383e39a3185df540048f62e4f9/fs/find-packages/src/index.ts#L7) and [Yarn](https://github.com/yarnpkg/berry/blob/13d5b3041794c33171808fdce635461ff4ab5c4e/packages/yarnpkg-core/sources/Workspace.ts#L2) use and we're aiming for compatibility with them.

benmccann:
Thanks @emmatown! I really appreciate all the reviews. I've updated the PR to stick with `fast-glob` for now. I'll see if pnpm and yarn might be willing to switch and perhaps we could revisit it later on if so

emmatown (approved):

Issue #219, opened 2024-08-12
`Packages` no longer serializable


benmccann:
In `get-packages` version 1 the returned `Packages` type was serializable. In version 2 it is not:
https://github.com/Thinkmill/manypkg/blob/045f7d7dabfa8a25b9d160661184a35eb4b5eaa7/packages/tools/src/Tool.ts#L116

`changesets` is stuck on version 1 as a result.

`changesets` is responsible for more than 80% of the usage of this package: https://packages.ecosyste.ms/registries/npmjs.org/packages/@manypkg%2Fget-packages/dependent_packages?kind=runtime&latest=true&order=desc&sort=downloads

Would you accept a PR to restore the serializability of the return value?

Andarist:
> Would you accept a PR to restore the serializability of the return value?

I'd be interested in your take on the problem. 

When v2 was released I didn't realize that this serializability was needed for Changesets 😢

PR #220, opened 2024-08-26
feat: serializable return values

benmccann:
closes https://github.com/Thinkmill/manypkg/issues/219

### DIFF
--- a/.changeset/chilly-bottles-eat.md
+++ b/.changeset/chilly-bottles-eat.md
@@ -2,4 +2,4 @@
 "@manypkg/find-root": minor
 ---
 
-feat: export DEFAULT_TOOLS
+Added `DEFAULT_TOOLS` export
--- a/.changeset/chilly-bottles-eat.md
+++ b/.changeset/chilly-bottles-eat.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/find-root": minor
+---
+
+feat: export DEFAULT_TOOLS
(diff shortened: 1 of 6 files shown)


packages/tools/src/Tool.ts
```diff
@@ -97,7 +97,7 @@ export interface MonorepoRoot {
   /**
    * The underlying tool implementation for this monorepo.
    */
-  tool: Tool;
+  tool: string;
```
Andarist, line 100: this is a breaking change, right?

benmccann: Yeah, I listed it as a major in the changeset.

.changeset/chilly-moles-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/tools": major
+---
+
+feat: serializable return values
```
Andarist, line 5:
```suggestion
Return values became serializable again. An identifier for the detected tool is returned instead of an instance of a `Tool` object.
```

.changeset/chilly-bottles-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/find-root": minor
+---
+
+feat: export DEFAULT_TOOLS
```
Andarist, line 5:
```suggestion
Added `DEFAULT_TOOLS` export
```

Issue #221, opened 2024-09-01
Large dependency graph


kachkaev:
:wave: folks! I’ve just tried `manypkg` instead of [`check-dependency-version-consistency`](https://www.npmjs.com/package/check-dependency-version-consistency) – looks great!

There is a minor internal issue I would like to bring up. Installing `@manypkg/cli@0.21.4` adds quite a lot of transient dependencies some of which are quite dated. An example would be `spawndamnit@2.0.0` that [has not been updated for six years](https://github.com/jamiebuilds/spawndamnit).

Because of that, the lock file gets quite polluted. Here is my diff after swapping `check-dependency-version-consistency` with `@manypkg/cli` (quite a lot of new stuff):

![Screenshot 2024-09-01 at 11 35 27](https://github.com/user-attachments/assets/575faf12-c592-49bc-a64b-dd3cf46c51e5)

Because the new dependency graph is quite big and parts of it are dated, there is a risk of bumping into security advisories that will be hard to address. It’d be great if the number of deps could be made smaller and libraries like `spawndamnit` could be replaced with something else, if possible.

Despite this small concern, great tool folks! I really like the simplicity of the DX you’ve created!

Andarist:
We could accept a PR swapping this dependency for a lighter alternative. I'm not sure what that alternative would be though.

VanTanev:
@Andarist I will open a PR with some replacements - some obvious candidates that jump out to me looking at the dependency graph https://npmgraph.js.org/?q=%40manypkg%2Fcli 
 - upgrading package-json will remove 21 deps
 - chalk -> picocolors
 - fast-glob -> tinyglobby
 - fs-extra is easy to remove

Andarist:
Please just make sure that you don't change the required node version in the process.

VanTanev:
Current node version requirement seems to be 14.18.0. Is this the version that should be targeted? It went EoL in the beginning of 2023, but if that's the minimum requirement, I'll maintain it.

Andarist:
I think it's preferred to maintain it within the current major line. It would be a braking change to change it. We can consider releasing a new major too but I don't quite have time to focus on it so it would be better to just keep the status quo for the time being.

VanTanev:
Hey @Andarist, do you prefer one big PR that removes/replaces all packages, or separate PRs per package?

Andarist:
Separate

riderx:
There is also js-yaml who is big for find-root package.

[...]

PR #224, opened 2024-10-13
[dependencies-replacement] Replace spawndamnit with tinyexec (8 fewer dependencies)

VanTanev:
First PR addressing #221.

https://github.com/tinylibs/tinyexec is used by `vitest`, so it should be a fairy safe bet.


packages/cli/src/index.ts
```diff
@@ -93,11 +93,14 @@ async function execCmd(args: string[]) {
   await Promise.all(
     packages.map((pkg) => {
       return execLimit(async () => {
-        let { code } = await spawn(args[0], args.slice(1), {
-          cwd: pkg.dir,
-          stdio: "inherit",
+        const proc = exec(args[0], args.slice(1), {
```
Andarist, line 96: Could u refactor the touched lines to this style (when possible)?
```suggestion
        const { exitCode } = await exec(args[0], args.slice(1), {
```

or at least to this if the `.exitCode!` is problematic:
```suggestion
        const result = await exec(args[0], args.slice(1), {
```

VanTanev: Unfortunately, the `exec()` constructor and the return type of the `await proc` are different types. 

`const proc = exec()` is an object that holds a "prepared execution" that you can attach streams/events/whatever on. It is also .then()-able, but that returns a different object...
`const result = await proc` is a Result type that only has `stdout` and `stderr`. It doesn't have `exitCode` (though I think it really should, there is an issue upstream, will see if I can open a PR for it).

The `proc.exictCode!` should be correct, because `exitCode` should always be set if we have awaited the `proc`.

As weird as the code looks right now, it is the only way to write it as far as I can see.

Andarist: Let's wait a little bit then before landing this. It would be great if `exitCode` thing would be improved upstream.

VanTanev: Oh, I was holding it wrong!

When the process has an non-zero exit code, an exception is thrown.

This means that if we don't read stdout/stderr, we can just do:
```
    await exec("yarn", args.slice(1), {
      nodeOptions: {
        cwd: matchingPackages[0].dir,
        stdio: "inherit",
      },
    });
 ```
 
 And we can completely remove the `ExitError` class, as `tinyexec` throws its own `NonZeroExitError`

Comment (VanTanev) [reply packages/cli/src/index.ts]: Merged the upstream change in `tinyexec`: https://github.com/tinylibs/tinyexec/pull/38

The PR is now good for re-review.

File: packages/cli/src/index.ts [hunk 2]
```diff
@@ -93,11 +93,13 @@ async function execCmd(args: string[]) {
...
-          cwd: pkg.dir,
-          stdio: "inherit",
+        const { exitCode } = await exec(args[0], args.slice(1), {
+          nodeOptions: {
+            cwd: pkg.dir,
+            stdio: "inherit",
+          },
         });
-        highestExitCode = Math.max(code, highestExitCode);
+        highestExitCode = Math.max(exitCode ?? 1, highestExitCode);
```
Andarist, line 102: why do we need to treat a nullish `exitCode` as a failure here?

VanTanev: Per `node:child_proccess.spawn()`, `exitCode` can be nullish in 2 cases:
 - The process failed to spawn.
 - The process was externally killed.
 
For our purpose, both of these mean that the underlying process never executed successfully, so we should tread a missing exit code as an error.

types/spawndamnit.d.ts
Andarist, line 1: 😍

PR #225, opened 2024-10-13
[dependencies-replacement] Replace fs-extra with node:fs/promises

VanTanev:
For #221 - removes 4 dependencies

### DIFF
--- a/.changeset/thick-panthers-smile.md
+++ b/.changeset/thick-panthers-smile.md
@@ -2,4 +2,4 @@
 "@manypkg/cli": minor
 ---
 
-Replace fs-extra with native node:fs/promises
+Replace `fs-extra` with native `node:fs/promises`
--- a/.changeset/thick-panthers-smile.md
+++ b/.changeset/thick-panthers-smile.md
@@ -1,0 +1,5 @@
+---
+"@manypkg/cli": minor
+---
+
+Replace fs-extra with native node:fs/promises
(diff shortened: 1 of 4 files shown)


.changeset/thick-panthers-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/cli": minor
+---
+
+Replace fs-extra with native node:fs/promises
```
Andarist, line 5:
```suggestion
Replace `fs-extra` with native `node:fs/promises`
```

VanTanev: @Andarist ill update the other PRs with this formatting too, so don't bother changing it there.

packages/cli/package.json
```diff
@@ -17,7 +17,6 @@
     "chalk": "^2.4.2",
     "detect-indent": "^6.0.0",
     "find-up": "^4.1.0",
-    "fs-extra": "^8.1.0",
```
Andarist, line 20: ugh, I've missed it but we might also want to drop this: https://github.com/Thinkmill/manypkg/blob/d0f4d926040bca29b1f2dd9f366e60854fb16118/package.json#L32

VanTanev: I'll add it to one of the next ones.

Wanted to add knip to catch stuff like this, but even though it's a dev dependency, it requires node 18.18. If we're fine with removing the engines field in the monrepo root?

Andarist: Hm, yee - the monorepo root doesn't have to maintain that node version compat. Although I guess having the `engines` there makes it easier to control stuff right now. But well, I'd still be OK with removing it from there and making sure it's in all other `package.json` of our publishable packages

PR #228, opened 2024-10-13
Add knip

VanTanev:
As per the discussion in https://github.com/Thinkmill/manypkg/pull/225#discussion_r1798465159 - remove `engines` from the root `package.json` and add `knip`. Also add CI task for `knip`. 

All published packages already have `engines` set at `node >= 14.18.0`.

Accidentally, this removed 4 dependencies from `@manypkg/cli`, because `find-up` dependency was unused.

References #221

### DIFF
--- a/.changeset/rude-geckos-happen.md
+++ b/.changeset/rude-geckos-happen.md
@@ -1,0 +1,7 @@
+---
+"@manypkg/get-packages": patch
+"@manypkg/tools": patch
+"@manypkg/cli": minor
+---
+
+Add `knip` and fix all issues raised by it.
--- a/.changeset/rude-geckos-happen.md
+++ b/.changeset/rude-geckos-happen.md
@@ -1,7 +1,5 @@
 ---
-"@manypkg/get-packages": patch
-"@manypkg/tools": patch
 "@manypkg/cli": minor
 ---
 
-Add `knip` and fix all issues raised by it.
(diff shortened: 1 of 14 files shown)


packages/cli/package.json
```diff
@@ -27,6 +26,7 @@
     "validate-npm-package-name": "^3.0.0"
   },
   "devDependencies": {
+    "@changesets/types": "^5.2.1",
```
Andarist, line 29: where do we depend on this? 🤔

.changeset/rude-geckos-happen.md
```diff
@@ -0,0 +1,7 @@
+---
+"@manypkg/get-packages": patch
+"@manypkg/tools": patch
+"@manypkg/cli": minor
+---
+
+Add `knip` and fix all issues raised by it.
```
Andarist, line 7: Adding `knip` is an internal change and the users (changesets target audience) don't care about this. Don't bother documenting what changes internally but please document the user-facing changes using changesets (like for example the removal of `find-up`)

packages/cli/package.json
VanTanev: https://github.com/Thinkmill/manypkg/blob/4a835ed40fe40b046810b60c00d918a5aa36dfc8/packages/cli/src/npm-tag.ts#L2

PR #181, opened 2023-06-10
Updated `package-json` dependency

fbartho:
Fixes: #180 

package-json went ESM-only, but we only use it in one place in upgrade.ts, and the error that was thrown told me to use a dynamic import, so I did. It was already an `async` context, so I think this is fine?

The tests pass!

Many yarn.lock updates though!

PR #231, opened 2024-10-14
[dependencies-replacement] Update package-json to latest, removing transitive dependencies

VanTanev:
References #221 

Upgrading `package-json` removes 21 dependencies. Necessitated changing node requirement to `>=18.18.0` in `@manypkg/cli`.
Then upgrading `semver` and `sembear` removes 2 dependencies, by making all 3 resolve to the same version of `semver`.

Total: 23 dependencies removed.

#### Additional changes

- Upgraded `@preconstruct/cli` and `@babel/*` dev dependencies, because I initially thought they were the cause of dynamic import not working with `preconstruct dev`, but it turned out that a config change was necessary in `babel.config.js`. I can revert this if you prefer, but it seems to work OK with the latest versions.
- Upgraded `fsevents` transitive dev dependency because the current version was broken on M1 Macs.


VanTanev:
Should I be squashing the changeset commit into the commit that has the actual changes?

packages/cli/src/upgrade.ts
```diff
@@ -96,12 +99,45 @@ export async function upgradeDependency([name, tag = "latest"]: string[]) {
...
+
+    const headers: Record<string, string> = {
+      accept:
+        "application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*",
+    };
+    if (authInfo) {
+      headers.authorization = `${authInfo.type} ${authInfo.token}`;
+    }
+
+    const response = await fetch(packageUrl, { headers });
```
Andarist, line 122: This is totally something I'd prefer not to have in the code 😢 I'm happy to offload this kind of code to dependencies just so I can stay ignorant of the requirements for fetching this thing.

At the very least, this should be moved to a separate function and some resources should be added in the comment to explain what's happening (I suspect most of this comes from the `package-json` package, if that's the case we should have proper license attribution for that package).

Is `package-json` maintained? Could PRs be open against that project to shrink its size?

VanTanev: It is maintained, but it's ESM-only and `node >=18.18.0` only.

The latest version is significantly smaller, because it uses `ky` instead of `got` for network requests

Andarist: The CLI package here is still on 0.x version and you have already scheduled some minor releases for it (which are effectively treated as breaking for 0.x packages). Let's bump the required node version for the CLI alone

VanTanev: What about `package-json` being ESM-only? I don't know of a clean solution, other than converting the cli package to ESM.

Andarist: Since this is a CLI you could also convert it to ESM but this will likely be more complicated given the test setup, tsconfigs and all. So the dynamic import parth forward (for now) is probably just simpler

VanTanev: Roger, will look into this!

VanTanev: OK, I needed to change `babel.config.js` to make `preconstruct dev` work. Other than that, using dynamic import() was absolutely the correct call, thanks!

Should be good for re-review, I've also updated the PR description.

.changeset/olive-tomatoes-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/cli": minor
+---
+
+Node requiremet raised to `node >= 18.18.0`. Upgraded `package-json`, `semver` and `sembear` dependencies.
```
Andarist, line 5: the updated dependencies should be one changeset file and the bumped node requirement should be another. Even though they are introduced by a single PR they will like 2 separate changes for the end user

babel.config.js
```diff
@@ -4,6 +4,8 @@ module.exports = {
       "@babel/preset-env",
       {
         targets: { node: 14 },
+        // necessary to preserve dynamic import of ESM-only modules
+        exclude: ["proposal-dynamic-import"],
```
Andarist, line 8: have you checked the build output? i'd expect that preconstruct flag I mentioned to be required too - although I never really used it myself so far

babel.config.js
```diff
@@ -4,6 +4,8 @@ module.exports = {
       "@babel/preset-env",
       {
         targets: { node: 14 },
```
Andarist, line 6: this should be updated now :p

Andarist: ah, maybe it shouldnt - since we are only bumping the node version for the CLI itself

babel.config.js
VanTanev: The preconstruct flag was already set, and the binary worked correctly with `yarn preconstruct build` but not with `yarn preconstruct dev` until the babel change was made.

.changeset/olive-tomatoes-provide.md
VanTanev: Thanks, I'm not used to writing changesets so this is very helpful!

babel.config.js
Andarist: ah cool, I see we are using it since https://github.com/Thinkmill/manypkg/pull/181

babel.config.js
VanTanev: I actually don't know `@preconstruct/cli`'s exact behavior here - does the build downlevelling use babel or TSConfig to decide what to output?

Andarist: Runtime files are exclusively~ transformed by Babel

> it seems like preconstruct build does not consider babel.config.js

It sounds like a configuration issue or a bug in Preconstruct. How did you come up with that conclusion? What I could try to repro locally and how I could observe the broken output?

VanTanev: - In `@manypkg/cli` upgrade `package-json@^10.0.1`.
 - `yarn install`
 - `yarn preconstruct dev`
 - `yarn manypkg check` # broken
 - `yarn preconstruct build`
 - `yarn manypkg check` # works

VanTanev: `yarn preconstruct dev` starts working only after you explicitly add `exclude: ["proposal-dynamic-import"],` to `babel.config.js`, while `yarn preconstruct build` works only if you have
```    
    "___experimentalFlags_WILL_CHANGE_IN_PATCH": {
      "keepDynamicImportAsDynamicImportInCommonJS": true
    }
```

VanTanev: Relevant preconstruct code is that the experimental flag adds a rollup plugin:

https://github.com/preconstruct/preconstruct/blob/54c412d6f10752e3e4d0aa04aced1c44958baee1/packages/cli/src/build/rollup.ts#L221C1-L222C79

So, babel seems to be completely ignored in that regard, and I don't see anything that would get the preset-env babel target in that file either.

VanTanev: It seems like the rollup setting that would control build output is `outputs.generatedCode`. It doesn't seem like `preconstruct` sets this explicitly, which probably means that the build uses the rollup defaults: 

https://github.com/rollup/rollup/blob/v2.79.2/src/utils/options/normalizeOutputOptions.ts#L346-L363

Andarist: nice github digging, my style - i love it :) The second link might be slightly inaccurate as it points to `master` an not to the tagged version but I doubt it matters here much (pro tip - you can hit `y` to get a permalink in the URL bar)

> It seems like the rollup setting that would control build output is outputs.generatedCode. It doesn't seem like preconstruct sets this explicitly, which probably means that the build uses the rollup defaults:

I'm slightly confused here given the plugin in Preconstruct that you linked to. But tbh this might just be the issue of not checking out locally how it behaves and what's the actual current output. Given you already looked into it - do you think that anything has to be fixed in Preconstruct itself? 

I'd treat the configuration issue with Babel for `preconstruct dev` purposes separately. This might not be something to fix in Preconstruct. So I'm primarily interested in the effect of all of this on `preconstruct build`

VanTanev: Yeah, the second link is to master, [but it's the same for version 2.79.2](https://github.com/rollup/rollup/blob/v2.79.2/src/utils/options/options.ts#L86-L92).

As for preconstruct... I don't think preconstruct is doing anything *wrong*, per se. It's not obvious that you need to configure 2 separate settings for `build` and `dev` to execute in the same way, but it does make some sense. (tangentially, `keepDynamicImportAsDynamicImportInCommonJS` would be easier to implement in `rollup@3`, because it's now [a separate setting that defaults to `true`](https://rollupjs.org/migration/#dynamic-import-in-commonjs-output), so `preconstruct` wouldn't need a custom plugin.)

On the more broad question of what the output of `preconstruct build` should be - there doesn't seem to be a way to configure `preconstruct` to output more modern JS. `ES2015` is fully supported on `node >= 10.15.0` (and `node 8` is missing only like 2 methods), but the rollup `es2015` preset isn't even the full `ES2015` standard. I think it would make sense to change `preconstruct build` output to use the [the es2015 rollup preset](https://github.com/rollup/rollup/blob/c9bd03d12e96c46122a0372d3bbe9b468cee57da/src/utils/options/options.ts#L79-L85)), as those parts should be `node >= 8.0.0` I think

Preconstruct could also consider enabling `keepDynamicImportAsDynamicImportInCommonJS: true` by default, but that requires `node >= 14.0.0` so it's a harder config change.

emmatown: The Rollup output options aren't really relevant, they're not for down-levelling, they only control a small amount of helper code that Rollup sometimes needs to emit. Updating the target in Babel will make `preconstruct build` output more modern JS

VanTanev: Ah. Apologies then, I was completely wrong.

babel.config.js
Andarist: I wonder if this would still be required if we'd update all of the Babel deps. Maybe we just have an outdated "stable features" list in the `@babel/preset-env` since we are using an old version?

packages/cli/package.json
```diff
@@ -31,6 +31,6 @@
     "strip-ansi": "^6.0.0"
   },
   "engines": {
-    "node": ">=14.18.0"
+    "node": ">=18.18.0"
```
Andarist, line 34: I think we could just use `"node": ">=18.0.0"`. 18.18.0 is a weird threshold, it's neither the current latest 18 nor it's the one that when 18 was marked as LTS for the first time (that was 18.12.0).

VanTanev: It is weird. I thought we needed that patch version for some reason, but looking at the changelog I'm completely wrong.

I think that we should be fine setting it at `"node": ">=18.0.0"`, I don't see a reason we would need the LTS specifically.

PR #232, opened 2024-10-15
[dependencies-replacement] Replace fast-glob with tinyglobby

VanTanev:
References #221 

Removes 15 dependencies from the graph.


VanTanev:
This one doesn't have a lot of tests attached to it. I would be a bit worried about changed behavior, possibly on windows. Not sure how best to test it - maybe add a windows CI path? Or, is windows generally considered supported?

Andarist:
A CI job for windows would be appreciated regardless - it might be good to PR it separately from this.

I'm similarly worried about the changed behavior but as long as this is a widely used package and since we'll be releasing a new 0.x minor with those changes... it should probably be fine.

packages/tools/package.json
```diff
@@ -9,9 +9,9 @@
   "license": "MIT",
   "main": "dist/manypkg-tools.cjs.js",
   "dependencies": {
-    "fast-glob": "^3.3.2",
     "jju": "^1.4.0",
-    "js-yaml": "^4.1.0"
+    "js-yaml": "^4.1.0",
+    "tinyglobby": "^0.2.9"
```
Andarist, line 14: hm, this change actually touches `@manypkg/tools` and not the CLI itself - that makes me way more worried about this change since `tinyglobby` might slightly differ in some cases from `fast-glob`

we might have to consider releasing v2 of this package some time in the future to support the upcoming Changesets v3 release, so maybe that would be the best time to revisit this? wdyt?

VanTanev: Yes, I was gonna raise that this change requires releasing new versions of `@manypkg/tools` and `@manypkg/get-packages`.

This change does lower the dependency count substantially (by the time everything else merges, it would be by about 30%). However I entirely understand that you don't want it rushed, so just let me know how to best support its eventual merge.

benmccann:
I just stumbled across this PR. I thought it'd be good to mention that nx has switched to tinyglobby and that lerna and pnpm may be on the verge of switching as well (there are PRs out for both that have gotten positive reactions from maintainers)

packages/tools/package.json
benmccann, line 14:
```suggestion
    "tinyglobby": "^0.2.12"
```

benmccann:
pnpm switched to tinyglobby awhile ago and lerna just merged their PR to switch to it as well

PR #234, opened 2024-10-17
Revert unnecessary change to babel config

VanTanev:
Turns out, @Andarist was correct in https://github.com/Thinkmill/manypkg/pull/231#discussion_r1804429503 - because babel was updated, this setting is not needed.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,8 +4,6 @@ module.exports = {
       "@babel/preset-env",
       {
         targets: { node: 14 },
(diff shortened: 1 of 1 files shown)


Andarist:
Hm, when was Babel updated though? 😅 I don't see any recent upgrades - so I wonder: why you have observed broken behavior before and why the change turned out to be unnecessary now?

VanTanev:
My guess is that my local `node_modules` got into a bad state somehow? I cannot replicate the error I was getting before :sob:

PR #236, opened 2024-11-04
In monorepo root, disallow dependencies in favor of devDependencies

### LINKED ISSUE #233: Should there be an option to force devDependencies in root package.json?
I think that there is a valid case to be made that manypkg should optionally enforce all dependencies be in `devDependencies` in the root.

We have a monorepo that needs to build docker containers. It has separate pnpm installs for the build step and the production step.

When it builds the production step, it uses [turbo prune](https://turbo.build/repo/docs/guides/tools/docker) and then `pnpm install --prod` - this will unfortunately also install all monorepo management tools defined in the root `package.json#dependencies`, needlessly increasing the docker image size.

Forcing root to use `devDependencies` instead would solve for this, and to some degree, I think it makes more intuitive sense if stuff like `typescript`, `prettier`, `lefthook` etc are dev dependencies instead of prod dependencies.

If there is interest from maintainers and guidance on desired implementation, I would be happy to contribute a PR.

VanTanev:
closes #233


VanTanev:
Anything in here that needs additional work?

kachkaev:
👋 @Andarist! What are your thoughts on getting this one in?

Andarist (approved):
It would be good to have @emmatown chime in too

emmatown (approved):

kachkaev:
Awesome, thanks for getting this in folks! I can't wait to try a new version and replace `dependencies` with `devDependencies` in the root `package.json` 😅

PR #241, opened 2025-04-21
Upgrade prettier to v3

bluwy:
Upgraded prettier to v3 as a semi-step before updating the rest of the dependencies.


.prettierrc
bluwy, line 1: I've moved it to a `.prettierrc` file so that we don't have to deal with the CJS/ESM stuff if/when we move to it later. Downside is we lose comments though but I'm not sure if it's crucial to keep them.

package.json
```diff
@@ -22,6 +22,9 @@
       "jest-watch-typeahead/testname"
     ]
   },
+  "resolutions": {
+    "prettier": "^3.5.3"
+  },
```
bluwy, line 27: I added this so that changesets uses prettier v3 instead of its included v2.

Andarist: Changesets should still prioritize the version used by the repo - so removing this should be fine

bluwy: Hmm I think you're right, but this does prevent two versions of prettier installed locally

Andarist: Hm, riiight. We could do this instead:
1. land [this PR](https://github.com/changesets/changesets/pull/1639) sometime this week (I have reviewing it on my TODO list for this week ;p)
2. release new Changesets v3
3. switch to that here
4. remove this line + land this PR :p

bluwy: I just removed the resolution field for now. I'm not sure we have to block this PR for that. It'd be nice to start making the dependency upgrades soon in this repo.

About that PR, it's not 100% perfect without upstream formatly changes, but I think it's also good enough for the interim if we want to merge that.

PR #244, opened 2025-04-26
Set engines field to node 20 and above

Andarist:
node 18 has its EOL literally in 4 days


bluwy:
If we set to node 20, will we also do the same for changesets later? I feel like it's a bit early to move to v20 even if it's EOL soon, as consumers might still want to stick or support node 18, but I guess that depends on the timeframe changesets v3 is out and stable.

Also, is it intentional that `gatsby-source-workspace` doesn't have a changeset?

bluwy (approved):
Ok with node 20 support if we want to directly extend to it.

Issue #238, opened 2025-02-26
Drop support for bolt?


benmccann:
https://www.npmjs.com/package/bolt - less than 2,000 downloads / week
https://github.com/boltpkg/bolt - last commit was 4 years ago

used here: https://github.com/Thinkmill/manypkg/blob/e6d32a3211cf12dc6c3b71ba7b3caa10259f9fc3/packages/find-root/src/index.ts#L9

I'd like to cut a new major for the tools package in https://github.com/Thinkmill/manypkg/pull/220, but this would be a different package

Andarist:
cc @steve-taylor

PR #245, opened 2025-04-26
Drop support for Bolt

Andarist:
closes https://github.com/Thinkmill/manypkg/issues/238


packages/get-packages/README.md
```diff
@@ -1,8 +1,8 @@
 # @manypkg/get-packages
 
-> A simple utility to get the packages from a monorepo, whether they're using Yarn, Bolt, Lerna, pnpm or Rush
+> A simple utility to get the packages from a monorepo, whether they're using Yarn, npm, Lerna, pnpm or Rush
```
bluwy, line 3: Is npm monorepos actually supported in the package? IIUC since it uses `package.json#workspaces` it may be incorrectly detected as yarn.

Andarist: good catch, https://github.com/Thinkmill/manypkg/pull/250

PR #246, opened 2025-04-26
chore(dev deps) - Upgrade all remaining dependencies

spanishpear:
Yarn v1 is frozen, and not receiving bugfixes or features anymore!

This PR migrates manypkg internally to use yarn 4 instead :) 

There should be no breaking changes to the usages.

Will have to fix some things in  CI 🕸️


package.json
```diff
@@ -41,7 +41,7 @@
     "prettier-plugin-packagejson": "^2.5.10",
     "typescript": "^5.3.2"
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
+  "packageManager": "yarn@4.9.1+sha512.f95ce356460e05be48d66401c1ae64ef84d163dd689964962c6888a9810865e39097a5e9de748876c2e0bf89b232d583c33982773e9903ae7a76257270986538",
```
spanishpear, line 44: SHA generated via `corepack up`https://github.com/nodejs/corepack/issues/231#issuecomment-1383129848

__fixtures__/basic-with-scripts/yarn.lock
spanishpear, line 1: We need a basic yarn lock, else the `yarn` scripts will all error with "this package isn't part of the yarn workspace of `manypkg`

packages/cli/src/__snapshots__/run.test.ts.snap
```diff
@@ -22,27 +17,17 @@ exports[`Run command should execute "package start" and exit with 1: stderr 1`]
 
 exports[`Run command should execute "package start" and exit with 1: stdout 1`] = `""`;
 
-exports[`Run command should execute "package-one start" and exit with 0: stderr 1`] = `
-"warning package.json: No license field
```
spanishpear, line 26: This error doesn't exist anymore in yarn berry - https://github.com/search?q=repo%3Ayarnpkg%2Fberry++%22No+license+field%22&type=code 

Was not fun to debug that 😆

packages/cli/src/run.test.ts
```diff
@@ -1,8 +1,10 @@
 import fixturez from "fixturez";
 import stripAnsi from "strip-ansi";
 import { exec } from "tinyexec";
+import { createRequire } from "node:module";
 
-const f = fixturez(__dirname);
+const f = fixturez(import.meta.url);
+const require = createRequire(import.meta.url);
```
spanishpear, line 7: Added in order to get tests running with ESM

tsconfig.json
```diff
@@ -3,9 +3,11 @@
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
     "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
-    "module": "Node16" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    //"erasableSyntaxOnly": true, - we need TS 5.8 first
+    "module": "nodenext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
+    "erasableSyntaxOnly": true,
     "allowImportingTsExtensions": true,
+    "skipLibCheck": true,
```
spanishpear, line 9: skipLibCheck to avoid typechecking external `.d.ts` files - that have errors in them - mostly due to gatsby/react/JSX upstream issues :(

package.json
```diff
@@ -13,35 +13,39 @@
     "postinstall": "preconstruct dev && NODE_OPTIONS=--experimental-strip-types packages/cli/bin.js check",
     "knip": "knip",
     "release": "preconstruct build && changeset publish",
-    "test": "jest",
+    "test": "NODE_OPTIONS='--disable-warning=ExperimentalWarning --experimental-vm-modules' jest",
```
spanishpear, line 16: tell jest to run with ESM as per https://jestjs.io/docs/ecmascript-modules

knip.json
```diff
@@ -6,5 +6,12 @@
...
   },
-  "ignore": ["__fixtures__/**"]
+  "jest": {
+    "config": ["package.json"],
+    "entry": [
+      "**/__tests__/**/*.ts",
+      "**/?(*.)+(test).ts",
+      "**/__fixtures__/**/*.[jt]s"
+    ]
+  }
```
Andarist, line 16: q: what do we need this for?

packages/cli/src/run.test.ts
```diff
@@ -35,7 +37,8 @@ describe("Run command", () => {
             cwd: f.find("basic-with-scripts"),
             env: {
               ...process.env,
-              NODE_OPTIONS: "--experimental-strip-types",
+              NODE_OPTIONS:
+                "--experimental-strip-types --disable-warning=ExperimentalWarning",
```
Andarist, line 41: does this make `stripNodeWarnings` obsolete?

packages/cli/src/run.test.ts
```diff
@@ -1,8 +1,10 @@
 import fixturez from "fixturez";
 import stripAnsi from "strip-ansi";
 import { exec } from "tinyexec";
+import { createRequire } from "node:module";
```
Andarist, line 4: We could replace this with https://www.npmjs.com/package/import-meta-resolve , but honestly using either is just equally fine given this is a test file anyway

packages/cli/src/test-helpers.ts
```diff
@@ -4,7 +4,7 @@
 //
 // Who can say? ¯\_(ツ)_/¯
 
-import { Package } from "@manypkg/get-packages";
+import type { Package } from "@manypkg/get-packages";
```
Andarist, line 7: We should use `erasableSyntaxOnly` + `verbatimModuleSyntax` to catch issues like this

Andarist: I see you have already enabled both 👍

tsconfig.json
```diff
@@ -3,9 +3,11 @@
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
     "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
-    "module": "Node16" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    //"erasableSyntaxOnly": true, - we need TS 5.8 first
+    "module": "nodenext" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
```
Andarist, line 6: Node16 (or Node18) is, kind of, a safer value

tsconfig.json
```diff
@@ -26,7 +28,8 @@
 
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
-    // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    "noImplicitOverride": true /* Ensure overriding members in derived classes are marked with an explicit override modifier. */,
+    "noImplicitAny": true /* Raise error on expressions and declarations with an implied 'any' type. */,
```
Andarist, line 32: this is already enabled through `strict: true`

Andarist:
Given @bluwy's migration to Vitest... I think I'd close this one and split the work into slightly smaller pieces. I'm in favor of migrating the repo to pnpm so migrating (temporarily) to yarn 4 doesn't bring much value. With that migration in - we should then bump all the dependencies.

knip.json
spanishpear: For some reason knip was failing to [automaticaly detect our jest usage](https://knip.dev/features/script-parser), and hence marking most of the test files as unused files.

tsconfig.json
spanishpear: oh, it is too!

spanishpear:
Makes sense @Andarist  - will probably wait for the decision on gatsby package removal as well :D

PR #247, opened 2025-04-26
Remove `@manypkg/gatsby-source-workspace`

bluwy:
Currently the package has a low download count, and removing it helps simplify some dependency upgrades we're doing in other PRs. So just a suggestion here to remove it.


package.json
```diff
@@ -41,6 +41,9 @@
   },
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
   "preconstruct": {
+    "packages": [
+      "packages/*"
+    ],
```
Andarist, line 46: why did this become required to fix the CI? 🤔

bluwy: It was `"packages/!(gatsby)*"` which meant all packages except the gatsby one. We still need the logic for all the packages and I accidentally remove it.

Andarist: Ah, OK - I looked at this single commit out of context ;p It makes sense 👍

emmatown (approved):

Issue #97, opened 2021-05-13
Does this work with npm?


reintroducing:
Hello,
We don't use yarn, rather npm, and for the most part this works. However, at times when I run `manypkg fix`, I get the following:
```
> manypkg fix

☔️ error Error: spawn yarn ENOENT
☔️ error     at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
☔️ error     at onErrorNT (node:internal/child_process:480:16)
☔️ error     at processTicksAndRejections (node:internal/process/task_queues:83:21) {
☔️ error   errno: -2,
☔️ error   code: 'ENOENT',
☔️ error   syscall: 'spawn yarn',
☔️ error   path: 'yarn',
☔️ error   spawnargs: []
☔️ error }
```

It seems that under the hood this wants to use yarn but since I don't have it installed its erroring. Is it possible to get away from this?

flybayer:
Yeah this is a unexpected problem.

Workaround is to manually fix things, which defeats the purpose of `manypkg fix`

steve-taylor:
NPM was a bit late to the monorepo party, so understandably it's currently unsupported. I hope this changes soon.

tomdavidson:
Would a PR using something like [yarpm](https://www.npmjs.com/package/yarpm) or [preferred-pm](https://www.npmjs.com/package/preferred-pm) be welcomed?

BenJackGill:
+1 for this.

I am having the same problems.

Andarist:
It would be great if somebody could locate where the error is coming from. This is likely an easy fix.

PR #250, opened 2025-04-27
Fixed an issue with projects using npm workspaces being recognized as yarn projects

Andarist:
see https://github.com/Thinkmill/manypkg/pull/245#discussion_r2061313856

### DIFF
--- a/.changeset/large-plants-sit.md
+++ b/.changeset/large-plants-sit.md
@@ -1,0 +1,7 @@
+---
+"@manypkg/get-packages": major
+"@manypkg/find-root": major
+"@manypkg/tools": major
+---
+
+Fixed an issue with projects using npm workspaces being recognized as yarn projects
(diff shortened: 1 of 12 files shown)


bluwy:
I think this should also address https://github.com/Thinkmill/manypkg/issues/97

Andarist:
I pushed an extra fix to make this happen. It still won't work for non-monorepos that are detected as "root" and mapped to `yarn` there. There is no point in using manypkg itself with a single-package repository though but this is certainly weird and wrong in the general sense.

packages/cli/src/utils.ts
```diff
@@ -18,6 +18,7 @@ export async function install(toolType: string, cwd: string) {
   const cliRunners: Record<string, string> = {
     bolt: "bolt",
     lerna: "lerna",
+    npm: "npm",
```
emmatown, line 21: i think the condition below for pnpm needs to apply for npm as well

emmatown (approved):
besides the comment, looks good

PR #255, opened 2025-06-23
Added ignore missing package name to get-packages

tzdesign:
Hi there,

we are also working with changesets, but in our current project there are packages that generate package.json files within the build process that do not have a name. 

These files are used for module types like:

```json
{
  "type": "module"
}
```

Unfortunately there is no way to ignore them, neither with you nor with changesets. I thought I'd do a PR, let me know if there's anything else I should change here.

When you release a new version, I will also make a PR in changesets project, so that you can pass on the value via the config.

Thanks in advance


emmatown:
I don't think this is something we should do, this seems like expected behaviour if you have your glob set to `packages/**` or etc. you should use `packages/*` or etc. and target the actual packages

Issue #257, opened 2025-07-22
manypkg fix changes all package.json files to be lf line endings on windows


jasekiw:
Like the title says, It converts all the line endings to lf which is really annoying when on windows because it marks all the files are changed even if there are no changes.

PR #260, opened 2025-08-08
fix: line endings are now respected on windows

jasekiw:
On windows all the package.json files would get their line endings changed to lf. This would cause all the package.json files to be marked as modified. This pr fixes the issue by detecting the line endings of the package.json file and normalizes them back what they originally were

Closes https://github.com/Thinkmill/manypkg/issues/257

### DIFF
--- a/.changeset/clever-feet-move.md
+++ b/.changeset/clever-feet-move.md
@@ -1,5 +1,5 @@
 ---
-"@manypkg/cli": minor
+"@manypkg/cli": patch
 ---
 
 Keep detected line endings flavor of `package.json` files on Windows when updating those files


packages/cli/src/utils.ts
Andarist, line 1: could you look into adding a test case for this? Ideally it should be an e2e test in: https://github.com/Thinkmill/manypkg/blob/fc5d5e0be9496841f4e528532697e741c86a5ed5/packages/cli/src/run.test.ts

jasekiw: I will look into that. I noticed I am getting extra output from yarn when running the tests and I'm trying to figure that out. Maybe it's because yarn outputs this only on windows?

image

jasekiw: @Andarist I added tests to the `run.test.ts` file. It checks both cases. Let me know if you rather have it in `fix.test.ts`

jasekiw:
@Andarist Does everything look good now that I've added tests?

jasekiw:
@Andarist just a reminder -  I made all the requested changes 😄

packages/cli/src/run.test.ts
```diff
@@ -49,3 +44,64 @@ describe("Run command", () => {
...
+
+function convertFileLineEndings(path: string, targetLineEnding: LineEndings) {
+  let file = fs.readFileSync(path, "utf8");
+  // detect mixed line endings
+  if (
+    file.includes("\r\n") &&
+    (file.match(/\r\n/g) || []).length !== (file.match(/\n/g) || []).length
+  ) {
+    throw new Error("mixed line endings in fixture file: " + path);
+  }
```
Andarist, line 80: I'd print a warning or something but format with `\n`. Throwing could be considered a breaking change.

Andarist: ah, this is just a test code - this is fine

.changeset/clever-feet-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/cli": minor
```
Andarist, line 2:
```suggestion
"@manypkg/cli": patch
```

Andarist:
I'm sorry, I missed the previous notification ;p I was not holding this PR a hostage, I swear! I appreciate you paying the ransom though /j 😉

jasekiw:
😆 I saw the projects you work on felt inclined to give something. I miss 99% of my GitHub notifications so I understand!

PR #266, opened 2026-02-23
fix(tools): support object-form workspaces in BunTool

MatthieuMordrel:
## Summary

Bun supports an object-form `workspaces` field in `package.json` for features like [workspace catalogs](https://bun.sh/docs/install/workspaces#workspace-catalogs):

```json
"workspaces": {
  "packages": ["apps/*", "packages/*"],
  "catalog": { "some-dep": "^1.0.0" }
}
```

`BunTool` only checked `Array.isArray(pkgJson.workspaces)`, so it missed monorepos using this object form. This caused `findRoot` to fall through to `RootTool`, treating the project as a single-package repo and skipping all workspace packages.

The fix mirrors `YarnTool`'s existing approach — it already handles both `string[]` and `{ packages: string[] }`.

### Changes

- **`BunTool.isMonorepoRoot` / `isMonorepoRootSync`** — now recognize `workspaces.packages` in addition to the array form
- **`BunTool.getPackages` / `getPackagesSync`** — extract globs from either `workspaces` (array) or `workspaces.packages` (object)
- **`BunPackageJSON` type** — updated to `string[] | { packages: string[] }`

## Test plan

- [x] Added `basic-bun-object-workspaces` fixture for `findRoot` tests
- [x] Added `bun-object-workspace-base` fixture for `getPackages` tests
- [x] All 93 tests pass (92 existing + 2 new, 1 skipped)


packages/tools/src/BunTool.ts
```diff
@@ -16,7 +16,7 @@ import {
 import { readJson, readJsonSync } from "./utils.ts";
 
 interface BunPackageJSON extends PackageJSON {
-  workspaces?: string[];
+  workspaces?: string[] | { packages: string[] };
```
Andarist: it would be nice to declare the catalog's type here - even if we don't need it now.

Andarist (approved):

Andarist:
Please add a changeset file documenting the change.

MatthieuMordrel:
Done, added the catalog type to BunPackageJSON and included the changeset. Thanks for the review!

.changeset/fix-bun-object-workspaces.md
```diff
@@ -0,0 +1,5 @@
+---
+"@manypkg/tools": patch
+---
+
+fix(tools): support object-form workspaces in BunTool to handle catalog usage
```
Andarist:
```suggestion
Support object-form workspaces in `BunTool` to handle `catalog` usage
```

